PR #45, opened 2015-01-01
Add env variables to pick installed driver

daleharvey:


vvo:
Thanks @daleharvey, published as selenium-standalone@2.44.0-3

PR #47, opened 2015-01-02
Callback notifying when selenium is ready to use

daleharvey:


daleharvey:
Wasnt particularly certain about whether it should maintain the var server = selenium() API as well as the callback one, I would get rid of the sync api as I always end up needing to go async at some point

vvo:
That's great, yes we do not need the sync version anymore, we can bump major when we remove it.

PR #48, opened 2015-01-03
Update start-selenium

brandon-collins:
start-selenium was broken after last commit because no callback was being passed.
added a noop callback


vvo:
thx

PR #49, opened 2015-01-04
update config settings

brandon-collins:
conf.js assumed hub url port of 4444
if -port 222 is passed as selenium server argument
the server will start but then get killed after 60s timeout
added simple argument parser specifically looking for
a port passed as selenium standalone argument


vvo:
Thanks again, I also did some changes to the start-selenium bin so that it does not uses the cb() for now.

This project needs some serious refactoring :D

PR #52, opened 2015-01-09
Split the install and start methods

daleharvey:
So users can programatically install which versions they require

### DIFF
--- a/install.js
+++ b/install.js
@@ -8,12 +8,6 @@ var request = require('request');
 
 var conf = require('./conf.js');
 
-async.series([
-  setup,
-  download,
-  chmodChromeDr.bind(null, conf.chromeDr.path)
-], end);
-
 function setup(cb) {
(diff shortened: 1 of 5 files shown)


daleharvey:
Should really be testing this properly, looking at the test setup now

install.js
```diff
@@ -161,3 +155,15 @@ function getIeDriverPlatform() {
...
+
+exports.install = function(callback) {
+  async.series([
+    setup,
+    download,
+    chmodChromeDr.bind(null, conf.chromeDr.path)
+  ], callback);
+}
+
+if (require.main === module) {
```
vvo: @daleharvey what does this does

daleharvey: If you run it from the command line, it calls that function, if its required then nothing happens

daleharvey:
Closed in favour of v3

PR #53, opened 2015-01-09
Give selenium more time to start

daleharvey:


vvo:
Sorry I am the one responsible for lowering the timeout :)

PR #54, opened 2015-01-10
v3

vvo:
@daleharvey Following our discussions and your proposal, I came up with this first v3 version, see the readme for API changes.


vvo:
@daleharvey now the code is easier to grasp, less redundancy and dirty tricks.

There's a clear separation between programmatic and command line.

We can add tests, features, easily (like .run :))

First tests seems to show fine behavior.

daleharvey:
Awesome, I took a look through and looks pretty great, :+1: from me

vvo:
published it as 3.0.0 on npm, will try on windows but I have no osx to try.

PR #56, opened 2015-01-16
Return the child process at the outset

nevir:
Gives us the ability to monitor (and log) output to aid debugging when something went wrong (the callback is too late for a lot of the good stuff).

### DIFF
--- a/lib/start.js
+++ b/lib/start.js
@@ -54,4 +54,6 @@ function start(opts, cb) {
 
     cb(null, selenium);
   });
+  
+  return selenium;
 }


vvo:
While it's a trivial change I am not sure it's a good idea to have both a callback based API also returning an object.

Can you precise the use case you had?

I am not fully against this, just want to be sure there's not a use case we can solve inside the library.

nevir:
Mostly I'm looking for a verbose mode - I'd like to display stdout/stderr to my users if they ask for it, or if an error occurs (e.g. retroactively).

nevir:
Here's my previous behavior (for v2 of selenium-standalone): https://github.com/Polymer/web-component-tester/blob/master/runner/steps.js#L88-103

vvo:
> Aside:  for your start checking in v

Kudos to @daleharvey for this. So simple I had not thought of it.. :D

vvo:
Ok so looking at your behavior, the missing features for you would be:

- listen for exit of selenium before it has started: it should be an error right? seems legit (#57)
- provide a way to get the selenium log, either by returning you the child_process immediately or provide a seleniumLogger function that will listen for stdin/stdout for you

What do you tihnk?

nevir:
Yeah, passing a function to capture log data should cover my needs; and it nicely matches your solution in #59!

A _mild_ case for returning the child process is that it provides a bit more control. (for example, I can kill it w/o waiting for it to completely start - say if a user ctrl-c's). But, not too worried about that one.

vvo:
> A mild case for returning the child process is that it provides a bit more control. (for example, I can kill it w/o waiting for it to completely start - say if a user ctrl-c's). But, not too worried about that one.

It could also be an option of the programmatic interface: exitOnSIGINT, exitOnSIGTERM, these would be used by the command line API.

Maybe that's too much yes :)

nevir:
Yeah, too much :P

PR #58, opened 2015-01-16
[readme] Add a link to the chromedriver host for version checking

nevir:


vvo:
Great to see you contributing but what is the added value of this statement at this particular position in the readme?

You want to say that the chromedriver is not stored on selenium servers that's it?

nevir:
Honestly, couldn't find a great place to put that link - but I definitely think you should have it somewhere on the readme to help people figure out what the latest version of it is. No real preference as to where!

PR #61, opened 2015-01-20
Support custom binaries download url

fengmk2:
Fixes https://github.com/cnpm/cnpmjs.org/issues/566

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,20 @@ selenium-standalone install --drivers.chrome.version=2.13
 selenium-standalone start --drivers.ie.arch=ia32
 ```
 
+### Custom binaries url
+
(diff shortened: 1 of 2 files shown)


vvo:
I will have a look today to see how I can plug this in the programmatic API with options rather than env variables, you will access it by flags on the command line, enough? @fengmk2

fengmk2:
yep, enough for me.

fengmk2:
@vvo any suggestions?

vvo:
@fengmk2 see https://github.com/vvo/selenium-standalone/blob/master/HISTORY.md#320-2015-01-23

enough?

fengmk2:
@vvo thanks! Your is better.

PR #68, opened 2015-02-05
Ensure files are downloaded once.

### LINKED ISSUE #50: install shouldnt redownload if it already exists

ctumolosus:
Follows suggested implementation where ETAG and MD5 are compared to determine whether remote and local files are identical. 

After installation, ZIP files are stored locally for future reference.

Fixes #50

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -1,6 +1,7 @@
 module.exports = install;
 
 var async = require('async');
+var crypto = require('crypto');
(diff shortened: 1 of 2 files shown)


vvo:
Very nice work!

Released with some fixes, `npm install selenium-standalone@4`

ctumolosus:
Awesome. Thanks!

PR #84, opened 2015-03-20
Updated readme with two ways to retrieve the selenium logs

Robinfr:


vvo:
Nice, I will just add another line about the cli command line tool already using the stdio inherit thing

vvo:
See https://github.com/vvo/selenium-standalone/blob/9d9022e74c9d19b667fe341c8dd8e765de703cfb/README.md#logging

PR #89, opened 2015-04-08
Adding support for parsing -hub argument and using accompanying hostname...

vdice:
Adding support for parsing `-hub` argument, if provided, and using accompanying hostname and port values from said argument instead of 'localhost' and '4444'.  If a `-port` arg is also supplied, this value will take precedence.

### DIFF
--- a/lib/check-started.js
+++ b/lib/check-started.js
@@ -1,7 +1,7 @@
 module.exports = checkStarted;
 
 var request = require('request').defaults({json: true});
-var URI = require('URIjs');
+var getSeleniumHub = require('./get-selenium-hub.js');
 
 function checkStarted(seleniumArgs, cb) {
   var retries = 0;
@@ -31,23 +31,4 @@ function checkStarted(seleniumArgs, cb) {
   }
 
   hasStarted();
-}
-
-function getSeleniumHub(seleniumArgs) {
-  var port = 4444;
-  var hostname = 'localhost';
-  var portArg = seleniumArgs.indexOf('-port');
-  var hubArg = seleniumArgs.indexOf('-hub');
-
-  if (hubArg !== -1) {
-    hubURI = new URI(seleniumArgs[hubArg + 1]);
-    hostname = hubURI.hostname();
-    port = hubURI.port();
-  }
-
-  if (portArg !== -1) {
-    port = seleniumArgs[portArg + 1];
-  }
-
-  return 'http://' + hostname + ':' + port + '/wd/hub/status';
 }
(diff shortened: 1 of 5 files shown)


vvo:
seems fine, what do you mean by 

> Would like to also wrap unit tests around the getSeleniumHub function in check-started.js but need a little help getting my understanding of mocha kickstarted...

? Programmatic usage seems to be fine, you did not break anything. Wonder how we could add a unit test for this.

vdice:
Thanks, @vvo.

My idea for a unit test, specifically around these changes in `getSeleniumHub`, would be to invoke `selenium-standalone start -- args` for the following cases:

args | expected return value of `getSeleniumHub` 
-------------|-------------
-port 7777 | http://localhost:7777/wd/hub/status
-hub http://foo:8888/wd/register | http://foo:8888/wd/hub/status
-hub http://foo/wd/register -port 9999 | http://foo:9999/wd/hub/status

(those are the main basic cases, but could potentially use more.  perhaps later we might want option of getting http/s scheme from hub argument as well, etc.)

vdice:
@vvo check out the latest commit w/ unit tests.  To cover under test, I broke `getSeleniumHub` out of `check-started.js`.  Let me know if there's an alternate way you would prefer...

vdice:
@vvo I finally realized what my problem was all along.  In some cases, I am running `selenium-standalone start -- -role hub` with a `-hub http://foo:4444/wd/hub/register` where the actual selenium hub was another docker container.  Therefore, I realized I did not want the code to check if the hub was started at all.  See what you think -- the latest commit just checks if `'node'` exists in seleniumArgs and if so, bypasses the check.  Does that seem reasonable?  Would there be a more preferable way to approach this?

vvo:
> Therefore, I realized I did not want the code to check if the hub was started at all
I do not understand why you would not want to check if the hub is up when started as role node?

vdice:
My reasoning is probably a unique case.  Basically, my selenium hub is a separate docker container, linked to the nodes in the form of add'l docker containers.  (So the grid itself will fail if Selenium hub doesn't start appropriately.)  I've reverted that last commit and kept my changes as just the check-started.js change + unit tests... with your approval we can merge.

robertfausk:
What's the status of this PR and what about adding some doc?:)

robertfausk:
And what about adding support for the ```-browser``` option?

```
start -- - -role node -browser browserName=safari -browser browserName=chrome -browser browserName=firefox -hub http://localhost:4444/grid/register
```

I think this would be very useful in combination with ```node```, ```role```, ```port``` and ```hub``` option.

vvo:
It was merged and releasesd, you can do a PR for the readme if needed.

robertfausk:
Here we go;)
I'll create a new issue for the ```browser``` option.

PR #98, opened 2015-05-14
update to better handle server config options

### LINKED ISSUE #97: crash leaves selenium running
every time I launch selenium I end up with the following error
```
Volumes/750GBHD/usr/local/lib/node_modules/selenium-standalone/bin/selenium-standalone:53
        throw err;
              ^
Error: Unable to connect to selenium
    at hasStarted (/Volumes/750GBHD/usr/local/lib/node_modules/selenium-standalone/lib/check-started.js:18:10)
    at Timer.listOnTimeout (timers.js:110:15)
```
it seems like the check is failing, however I can still connect to selenium running on port 4444 and get the web based console

I believe it has to do with starting as a hub or a node
selenium-standalone start -- -role node -browser browserName=firefox,maxInstances=1 -browser browserName=chrome,maxInstances=1
or
selenium-standalone start -- -role hub

by doing this it changes it so that ttp://localhost:4444/wd/hub/status will return a 500

plessbd:
this fixes #97


lib/check-started.js
```diff
@@ -1,16 +1,14 @@
...
 
 function checkStarted(seleniumArgs, cb) {
   var retries = 0;
-
-  var hub = getSeleniumHub(seleniumArgs);
-
-  var retryInterval = 200;
+  var hub = getSeleniumStatusUrl(seleniumArgs);
   // server has one minute to start
-  var maxRetries = 60 * 1000 / retryInterval;
```
vvo: no need to microptimize this, otherwise you have to do the computation to know how much time you have to start

lib/check-started.js
```diff
@@ -25,10 +23,9 @@ function checkStarted(seleniumArgs, cb) {
...
         return;
       }
-
       cb(null);
     });
   }
-
-  hasStarted();
+  // give java a second to start before checking the status
+  setTimeout(hasStarted, 1000);
```
vvo: Since there's already a retry interval, why would we want to setTimeout this?

vvo:
Looks very good @plessbd, small changes and we are good.

plessbd:
Updated.
I put the setTimeout just to get rid of the noise made by some debugging I put in.  It actually starts the check for selenium before java has had time to start selenium.

vvo:
> I put the setTimeout just to get rid of the noise made by some debugging I put in. It actually starts the check for selenium before java has had time to start selenium.

You are right, I will put it back

Issue #104, opened 2015-06-11
Callback in selenium.start() called twice when process already exists


binarykitchen:
Looks like the callback in `selenium.start()` is being called twice when Selenium's Java process already exists. I mean

```coffee
  selenium.start({spawnOptions: {stdio: 'ignore'}}, (err, seleniumProcess) ->
    console.log '<------------------ selenium started'
    if err
      cb(err)
    else
      # do something else
  )
```

To reproduce, make sure you already have the Selenium process running and then run the above: you will see that the console output is printed twice and an error is passed onto the callback on the first run.

The first error says something like:

```
[12:54:33] 'selenium' errored after 309 ms
[12:54:33] Error: Selenium exited before it could start
  at ChildProcess.errorIfNeverStarted (/home/michael-heuberger/Development/.../node_modules/selenium-standalone/lib/start.js:92:10)
  at emitTwo (events.js:87:13)
  at ChildProcess.emit (events.js:172:7)
  at Process.ChildProcess._handle.onexit (internal/child_process.js:199:12)
```

When the callback is called on the second time, error object is null and I have a `seleniumProcess` but it looks weird:

```
{ domain: null,
  _events: {},
  _eventsCount: 0,
  _maxListeners: undefined,
  _closesNeeded: 1,
  _closesGot: 1,
  connected: false,
  signalCode: null,
  exitCode: 1,
  killed: false,
  spawnfile: '/usr/bin/java',
  _handle: null,
  spawnargs: 
   [ '/usr/bin/java',
     '-jar',
     '/home/michael-heuberger/Development/.../node_modules/selenium-standalone/.selenium/selenium-server/2.45.0-server.jar',
     '-Dwebdriver.chrome.driver=/home/michael-heuberger/Development/.../node_modules/selenium-standalone/.selenium/chromedriver/2.15-x64-chromedriver' ],
  pid: 20472,
  stdin: null,
  stdout: null,
  stderr: null,
  stdio: [ null, null, null ] }
```

I expect just the error to be thrown and the callback not to be called again.

vvo:
Could you try to open a pull request for this?

binarykitchen:
Here you go https://github.com/vvo/selenium-standalone/pull/105

dtothefp:
@binarykitchen @vvo I agree this is an issue but something like below works for me. Would be nice to have the error messages updated as @binarykitchen suggests

```js
import {join} from 'path';
import selenium from 'selenium-standalone';
[...]

PR #105, opened 2015-06-12
Fix for issue #104:

binarykitchen:
- Do not callback twice when process already exists
- Error message more clear
- Bump some packages

### DIFF
--- a/lib/start.js
+++ b/lib/start.js
@@ -65,6 +65,7 @@ function start(opts, cb) {
       return;
     }
 
+    var neverStarted = false;
     var selenium = spawn(opts.javaPath, args, opts.spawnOptions);
 
     // Add empty handler to stdout and stderr so the buffers can be flushed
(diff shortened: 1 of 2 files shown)


vvo:
Good, could you only include the necessary code change without any stylistic, formatting change?

Also, no need to move the export.

Thanks

binarykitchen:
well, i do not support these styles but have reverted that for you - otherwise all good. thanks for merging.

vvo:
The more I think about it the more I thinkg that we should throw/error in the case a selenium server is already running.

Otherwise this selenium instance would be unusable right?

binarykitchen:
No @vvo, the instance can be then reused right away - works fine from my gulp task. It is usable when already run. No need to throw an error.

vvo:
so you have a running selenium you start another one but then you still use the first instance?

binarykitchen:
no - that gulp task is a mocha test runner. sometimes it crashes when an error occurs and doesn't shut down selenium at all. then, imagine this: i fix the code and want to run the mocha tests again and then it complains that selenium is already running. in that case i have to `kill -9` the selenium process form the console and start the gulp task again.

that's why the above patch of mine will help and make it a bit easier for the developer. throwing an error doesn't help and makes it probably worse.

vvo:
> sometimes it crashes when an error occurs and doesn't shut down selenium at all.

so this is the real issue then?

vvo:
if that's fixable

binarykitchen:
there is no issue with my scenario, it's justified when the other gulp task crashes - it can happen and is not related to this issue. all i am saying is that the selenium child process is still running and not cleaned up which can happen in my scenario.

vvo:
4.4.3

PR #110, opened 2015-07-08
Update default versions

olizilla:
```sh
selenium-standalone install --version=2.46.0 --drivers.chrome.version=2.16
```
works fine, so bumping the default versions to latest.


vvo:
Thanks! released as 4.5.3

```sh
npm install selenium-standalone@4.5.3
```

Where do you use this project @olizilla?

olizilla:
Just tried it out on a new https://github.com/tableflip project, so I could get https://github.com/nightwatchjs/nightwatch with minimum faff, and it worked a charm. 
Thanks for making it!

PR #124, opened 2015-08-22
Optional Chrome and IE drivers. Fixes #123.

### LINKED ISSUE #123: Optional Chrome and IE drivers.
It'd be better if this library did not require an install of Chrome and IE drivers.

adjohnson916:
This makes Chrome and IE drivers optional. The default behavior remains (for backwards compatibility) to include them in both `install` and `start`, but they can be disabled as follows:

```js
{
  drivers: {
    chrome: false,
    ie: false
  }
}
```


vvo:
Thanks!

PR #129, opened 2015-09-13
Add tip about killing the server in the background

silvenon:
Let me know if I got it right, I'm no shell guru :sweat_smile:


vvo:
WFM :)

PR #135, opened 2015-09-25
Update package name and version of urijs

rocketsue:
* name changed from URIjs to urijs
* increase version number from 1.15.0 to 1.16.1


oliviertassinari:
The build fail here... We also need to renme `URIjs` to `urijs` in the source code

PR #138, opened 2015-09-30
Add install `--silent` option

ArnaudRinquin:
It'll allow transparent installation using npm scripts, like in the following
`package.json` example:

```json
{
  "scripts":{
    "install":"echo 'Installing selenium...'; selenium-standalone install --silent"
  }
}
```


vvo:
Great!

PR #148, opened 2015-11-16
Process can incorrectly terminate when running as a grid node

stephenash:
When `selenium-standalone` is run with only the `-role node` mode, eventually the process will die because the server status URL is returning an error. Example:

screen shot 2015-11-13 at 11 08 41

Looking at the only API documentation that I could find [0] and my own selenium test servers it appears that the `/wd/hub/status` API will always fail when called against the hub, but passes when called against a node. This PR changes to check against the `/grid/api/hub` API if running in a grid hub role (API to use was gotten from [0] and the selenium unit tests[1]). When running in a standalone mode or grid node, continue to use the existing API for checking status.

There was also some existing logic that tried to use the `-hub` parameter to decide what host the status check should query. It seems that the `checkStarted()` wants to know if the local process was started successfully, and not the status of the remote hub. There is now a new (unused) function `getRemoteSeleniumHubStatusUrl()` that can get the status URL for a remote hub if in the future there is a need to check for both the local process and the remote hub running.

---
[0] https://github.com/nicegraham/selenium-grid2-api/blob/master/README.md#wdhub
[1] https://github.com/SeleniumHQ/selenium/blob/master/java/server/test/org/openqa/grid/internal/StatusServletTests.java#L284

### DIFF
--- a/lib/get-selenium-status-url.js
+++ b/lib/get-selenium-status-url.js
@@ -9,6 +9,9 @@ function getSeleniumStatusUrl(seleniumArgs) {
   var hubArg = seleniumArgs.indexOf('-hub');
   var roleArg = seleniumArgs.indexOf('-role');
 
+  var nodeStatusAPIPath = '/wd/hub/status';
+  var hubStatusAPIPath = '/grid/api/hub';
+
   if (hubArg !== -1) {
     hubURI = new URI(seleniumArgs[hubArg + 1]);
     hostname = hubURI.hostname() || hostname;
@@ -21,10 +24,10 @@ function getSeleniumStatusUrl(seleniumArgs) {
         if(hubArg === -1){
           port = portArg !== -1 ? seleniumArgs[portArg + 1] : 5555;
         }
-        return 'http://' + hostname + ':' + port + '/wd/hub/status';
+        return 'http://' + hostname + ':' + port + hubStatusAPIPath;
       case 'hub':
         port = portArg !== -1 ? seleniumArgs[portArg + 1] : 4444;
-        return 'http://localhost:' + port + '/api/grid/hub/';
+        return 'http://localhost:' + port + hubStatusAPIPath;
     }
   }
 
@@ -32,5 +35,5 @@ function getSeleniumStatusUrl(seleniumArgs) {
     port = seleniumArgs[portArg + 1];
   }
 
-  return 'http://' + hostname + ':' + port + '/wd/hub/status';
+  return 'http://' + hostname + ':' + port + nodeStatusAPIPath;
 }
--- a/lib/get-selenium-status-url.js
+++ b/lib/get-selenium-status-url.js
@@ -19,8 +19,10 @@ exports.getSeleniumStatusUrl = function(seleniumArgs) {
   var processType = this.getRunningProcessType(seleniumArgs);
   var portArg = seleniumArgs.indexOf('-port');
   var port = (portArg !== -1) ? seleniumArgs[portArg + 1] : undefined;
+  var hostArg = seleniumArgs.indexOf('-host');
+  var host = (hostArg !== -1) ? seleniumArgs[hostArg + 1] : 'localhost';
 
-  var statusURI = new URI('http://localhost');
+  var statusURI = new URI('http://' + host);
   var nodeStatusAPIPath = '/wd/hub/status';
   var hubStatusAPIPath = '/grid/api/hub';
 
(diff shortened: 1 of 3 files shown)


lib/get-selenium-status-url.js
```diff
@@ -1,36 +1,71 @@
...
-      case 'node':
-        if(hubArg === -1){
-          port = portArg !== -1 ? seleniumArgs[portArg + 1] : 5555;
-        }
-        return 'http://' + hostname + ':' + port + '/wd/hub/status';
-      case 'hub':
-        port = portArg !== -1 ? seleniumArgs[portArg + 1] : 4444;
-        return 'http://localhost:' + port + '/api/grid/hub/';
-    }
+  var statusURI = new URI('http://localhost');
```
vvo: What I do not get is why it would always be "localhost" here? Can't the hub be started and only listen on another interface?

vvo:
Seems like you put a lot of good work here, I have a couple of questions

lib/get-selenium-status-url.js
```diff
@@ -1,36 +1,71 @@
...
   if (portArg !== -1) {
-    port = seleniumArgs[portArg + 1];
+    statusURI.port(seleniumArgs[portArg + 1]);
   }
 
-  return 'http://' + hostname + ':' + port + '/wd/hub/status';
+  return statusURI.toString();
+}
+
+exports.getRemoteSeleniumHubStatusUrl = function(seleniumArgs) {
```
vvo: If this current function is unused, when would it be used? Can you make it so that we already use it, or at least put a comment on top of it stating what's relevant to say

lib/get-selenium-status-url.js
stephenash: The only way that I can see the hub binding on another IP address or hostname is by using the `-host` parameter. The old code looked at the `-hub` parameter which had no bearing on the binding.

```shell
$ java -jar 2.47.1-server.jar -role hub -hub http://alias:4567/grid/register
...
2015-11-20 11:22:58.075:INFO:osjs.AbstractConnector:Started SocketConnector@0.0.0.0:4444
11:22:58.076 INFO - Nodes should register to http://153.32.222.55:4444/grid/register/
11:22:58.076 INFO - Selenium Grid hub is up and running
```

```shell
$ java -jar 2.47.1-server.jar -role hub -host alias -port 4567
...
2015-11-20 11:14:22.961:INFO:osjs.AbstractConnector:Started SocketConnector@alias:4567
11:14:22.964 INFO - Nodes should register to http://alias:4567/grid/register/
11:14:22.964 INFO - Selenium Grid hub is up and running
```

I didn't know about the `-host` parameter before, so I'll add support for that instead of hardcoding localhost.

---
Found the list of supported command line arguments for the `node` and `hub` roles in https://github.com/SeleniumHQ/selenium/blob/selenium-2.48.2/java/server/src/org/openqa/grid/common/defaults/NodeOptions.json and https://github.com/SeleniumHQ/selenium/blob/selenium-2.48.2/java/server/src/org/openqa/grid/common/defaults/HubOptions.json

stephenash: > I'll add support for that instead of hardcoding localhost.

Added with 64206e5

lib/get-selenium-status-url.js
stephenash: Originally my thinking was that on startup you might want to check that you are able to connect to the local process, and also check to see if you could connect to the remote hub (if there is one). If either of those connections cannot be made, then startup would be considered a failure. Since that is a change in behavior from the old code, I didn't added the hub check and only left the unused function.

There is probably a better second level check to be making instead of "is the remote hub responding". Something like "is the remote hub responding and is this node connected to said hub". I don't immediately see that type of API in the docs I found, so I'm going to back out the unused code instead (a244c49).

stephenash:
@vvo Fixed up the PR based on your questions. Ready for your review again.

vvo:
I think we are good @stephenash, I will also add you as a contributor with rights on github and npm. Very good work so I am sure you will be able, if you want, to handle updates/review :)

vvo:
@stephenash do you have an npm username?

stephenash:
@vvo Just signed up for npmjs with the same username: `stephenash`. Thanks!

vvo:
```
> npm owner add stephenash
+ stephenash (selenium-standalone)
```

always work with PR and follow http://semver-ftw.org/, thats it ;)

PR #150, opened 2015-11-19
Check if we can read and execute before chmoding.

btelles:
This chmod may not be necessary in some cases, and it breaks our ability to test in Docker. I'm sure I got some idioms wrong, and would be happy to fix them. Just let me know what to do.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -241,7 +241,14 @@ function basePath(fullPath) {
 }
 
 function chmodChromeDr(where, cb) {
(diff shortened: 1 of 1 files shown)


vvo:
Seems a nice addition. Only thing being that fs.access is not available in node 0.10. Wonder if we need to still support it..

vvo:
What do you think?

vvo:
Let forget about node 0.10, really too old.

Thanks!

PR #151, opened 2015-11-23
Update README.md

nekoyama:
Got the following error when using the code in the Programmatic API example : 

[12:15:28] 'selenium' errored after 334 ms
[12:15:28] Error: Could not request headers from http://selenium-release.storage
.googleapis.com/2/IEDriverServer_x64_2.45.zip:
    at Request.<anonymous> (c:\projects\Git\preciosita\node_modules\selenium-sta
ndalone\lib\install.js:289:17)

Yeah, only because I decided not copy paste the defaults just right below...


vvo:
Thanks!

PR #155, opened 2015-12-01
Use HTTPS URLs when possible.

### LINKED ISSUE #152: Should fetch executables over HTTPS or require matching checksums
It looks like selenium-standalone fetches Selenium and the WebDriver plugins over URLs like http://selenium-release.storage.googleapis.com/2.47/selenium-server-standalone-2.47.1.jar, which allows any network attacker to replace the binaries I'm about to execute. You should either fetch over HTTPS, or make your users pass in an expected hash of the download, which you should verify before unpacking it.

jyasskin:
Fixes #152.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -70,7 +70,7 @@
 
 # 4.1.0 (2015-02-06)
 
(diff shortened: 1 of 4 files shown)


vvo:
released as 4.8.0 with other fixes

PR #160, opened 2016-02-01
feat(upgrades): upgrade default versions

sublimino:
Selenium version `2.50.1` and IE version `2.50.0` via http://www.seleniumhq.org/download/
Chromedriver version `2.21` via https://chromedriver.storage.googleapis.com


vvo:
THX :)

PR #165, opened 2016-02-23
Do not merge the install/start options with the default options.

### LINKED ISSUE #128: start(opts, cb) is requiring drivers not in options
I'm using this code to start selenium server from Gulp
```js
  selenium.start({
    drivers: {
      chrome: {
        version: '2.15',
        arch: process.arch,
        baseUrl: 'http://chromedriver.storage.googleapis.com'
      }
    }
  }, function(err, childProc){
    console.log(err);
    child = childProc;
    done();
  })
```

The error I'm receiving says it's missing IE Driver, which I don't want to use and have no included in my start options.

```
[Error: Missing D:\dev\projects\gamedev-website\node_modules\selenium-standalone\.selenium\iedriver\2.47.0-x64-IEDriverServer.exe]
```

Install Script
```js
//Install selenium server for use with Protractor End to End testing
selenium.install({
  // check for more recent versions of selenium here:
  // http://selenium-release.storage.googleapis.com/index.html
  version: '2.47.1',
  baseURL: 'http://selenium-release.storage.googleapis.com',
  drivers: {
    chrome: {
      // check for more recent versions of chrome driver here:
      // http://chromedriver.storage.googleapis.com/index.html
      version: '2.15',
      arch: process.arch,
      baseURL: 'http://chromedriver.storage.googleapis.com'
    },
    ie: {
      // check for more recent versions of internet explorer driver here:
      // http://selenium-release.storage.googleapis.com/index.html
      version: '2.45',
      arch: process.arch,
      baseURL: 'http://selenium-release.storage.googleapis.com'
    }
  },
  logger: function(message) {
    console.log(message);
  },
  progressCb: function(totalLength, progressLength, chunkLength) {

  }
}, function(err){});
```

molily:
Fixes #128.

Let me know if this makes any sense. I tried to add a meaningful test, but `ChildProcess.spawnargs` is the only way I know to peek into the internals. It’s only available in newer Node versions. I’ve tested successfully with 0.10, 0.12 and 5.5, but only 5.5 can perform the `spawnargs` check.

This is an API-breaking change in terms of semver AFAICS. Existing code might depend on the fact that the default options are always merged in. Was there a specific reason to do this in the first hand?

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -3,7 +3,6 @@ module.exports = install;
 var async = require('async');
 var crypto = require('crypto');
 var fs = require('fs');
-var merge = require('lodash').merge;
 var mkdirp = require('mkdirp');
 var path = require('path');
 var request = require('request');
@@ -36,7 +35,9 @@ function install(opts, cb) {
 
   opts.progressCb = opts.progressCb || noop;
 
-  opts.drivers = merge(defaultConfig.drivers, opts.drivers || {});
+  if (!opts.drivers) {
+    opts.drivers = defaultConfig.drivers;
+  }
 
   logger('----------');
   logger('selenium-standalone installation starting');
--- a/lib/start.js
+++ b/lib/start.js
@@ -1,6 +1,5 @@
 module.exports = start;
 
-var merge = require('lodash').merge;
 var spawn = require('child_process').spawn;
 
 var which = require('which');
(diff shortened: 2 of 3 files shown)


vvo:
> Was there a specific reason to do this in the first hand?

Not that I remember, just an habit I guess.

Your solution is good and safer.

I will bump major (and add you as a collaborator on both github and npm).

vvo:
Sorry @molily I was wrong, there was a reason to do this merge: the README says that you can switch versions from the command line or even just programmatically.

Let's say you want all the defaults but just upgrade the chrome driver version, you can do it by just giving drivers.chrome.version = .. and everything else will be merged. You can trigger a bad behavior by doing that in your commit.

For this PR to be more complete:
- we need to compute the baseURL and arch properties when they are not given. This should solve our issues here
- we need to merge default parameters on the command line: https://github.com/vvo/selenium-standalone/blob/master/bin/selenium-standalone

What do you think?

PR #167, opened 2016-02-24
Merge drivers with default options, but only those specified.

molily:
Still allows to overwrite specific options on the CLI.

Second attempt :) Hope this addresses the points mentioned in
https://github.com/vvo/selenium-standalone/pull/165#issuecomment-188162273

See #165 and #128.

### DIFF
--- a/bin/selenium-standalone
+++ b/bin/selenium-standalone
@@ -2,8 +2,10 @@
 
 var minimist = require('minimist');
 var which = require('which');
+var merge = require('lodash').merge;
 
 var selenium = require('../');
+var defaultConfig = require('../lib/default-config');
 
 which('java', function javaFound(err, javaPath) {
   if (err) {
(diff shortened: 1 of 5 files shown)


lib/install.js
```diff
@@ -26,18 +27,25 @@ function install(opts, cb) {
...
   if (!opts.version) {
     opts.version = defaultConfig.version;
   }
 
-  if (!opts.baseURL) {
-    opts.baseURL = defaultConfig.baseURL;
+  if (opts.drivers) {
+    // Merge in missing driver options for those specified
+    opts.drivers = mapValues(opts.drivers, function(config, name) {
+      return merge({}, defaultConfig.drivers[name], config);
```
vvo: That's the cool trick! Congrats! And thanks a lot

bin/selenium-standalone
```diff
@@ -34,6 +36,9 @@ which('java', function javaFound(err, javaPath) {
     return prev;
   }, {});
 
+  // Merge in default options
+  options = merge({}, defaultConfig, options);
```
vvo: awesome

vvo:
released as 5.0.0

PR #172, opened 2016-03-22
Update default-config.js

plessbd:
New version of firefox has issues with older selenium server see https://github.com/SeleniumHQ/selenium/issues/1851.  I have updated all drivers to the newer versions as well


vvo:
Thanks!

PR #183, opened 2016-06-10
Remove deprecated graceful-fs@3.x dependency

Sanjo:
Replaces unmaintained unzip dependency with yauzl.
This removes the indirect dependency to the deprecated graceful-fs@3.x.


Sanjo:
The test that fails is the one that also fails on master.

vvo:
Thanks!

PR #187, opened 2016-06-11
Add support for Firefox Marionette Driver

### LINKED ISSUE #181: Firefox 48+ will require new Marionette driver
https://github.com/SeleniumHQ/selenium/issues/2110#issuecomment-224413609

It appears Firefox 47 is the last release that works out of the box with Selenium, and future releases will require the [Marionette Driver](https://developer.mozilla.org/en-US/docs/Mozilla/QA/Marionette/WebDriver)

azakus:
Required for Firefox 47+, supported for Firefox 45+

`geckodriver` executable is distributed as a .gz file for linux and osx,
so I've added a new gunzip workflow similar to the unzip workflow

Fixes #181

### DIFF
--- a/bin/selenium-standalone
+++ b/bin/selenium-standalone
@@ -14,7 +14,7 @@ which('java', function javaFound(err, javaPath) {
   }
 
   var argv = minimist(process.argv.slice(2), {
-    string: ['version', 'drivers.chrome.version', 'drivers.ie.version']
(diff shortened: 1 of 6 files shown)


azakus:
FYI, new driver requires `'marionette': true` in the selenium capabilities.

vvo:
Thanks a lot

PR #193, opened 2016-07-01
Update selenium to 2.53.1

jmalonzo:
This version should fix the issue with Firefox 47 and onwards as per https://github.com/seleniumhq/selenium/issues/2110#issuecomment-229684461


vvo:
Thanks!

PR #196, opened 2016-07-07
Add possibility to configure proxy via parameter 'proxy' like

### LINKED ISSUE #174: Proxy Hook
I have been looking into this project as a possible solution for downloaded selenium and specific drivers for my enterprises web applications. But one of the killer is the inability to use this tool behind an enterprise firewall. What is the possibility of adding proxy support as a configurable set of options that can be passed to the request module. Even the ability to download over http instead of https would be a nice start.

Additional Note:
Environmental variables are not an option where I work as all variables a flushed on a daily bases with fresh images nightly. This is a security thing and we can not change it.

alcaneo:
Fix #174

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -19,6 +19,7 @@ function install(opts, cb) {
   var progress = 0;
   var startedRequests = 0;
   var expectedRequests = 4;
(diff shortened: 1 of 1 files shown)


vvo:
Thanks!

PR #201, opened 2016-08-09
Make error message more descriptive.

kradical:
First thank you for your work on this project! It's really helpful. I updated the error message because this sort of thing would have prompted me to check what version of java the server was trying to start on (machine has java8 but was trying to start on java6). I know there is a couple issues open about this error message so hopefully this can close 1-2 of them.


vvo:
Thanks!

Issue #202, opened 2016-08-09
Support classpath


existdissolve:
Any plans to support classpath? I think this would be a nice feature, especially for supporting things like custom servlets and proxies. The basic idea, I think, would be similar to what is done here: 

https://github.com/jhoblitt/puppet-selenium

> classpath: Optional classpath to pass to Selenium Server Node at startup. If this array is empty Selenium will be started with '-jar /path/to/selenium.jar', if not empty Selenium will be started with '-cp /path/to/selenium.jar: org.openqa.grid.selenium.GridLauncher' See http://www.seleniumhq.org/docs/07_selenium_grid.jsp#customizing-the-grid

Thanks for your consideration!!

stephenash:
I believe that should be supported already through the [cli](https://github.com/vvo/selenium-standalone#command-line-interface)

```bash
selenium-standalone start -- -cp /custom/lib1.jar
```

or the [api](https://github.com/vvo/selenium-standalone#seleniumstartopts-cb)

```javascript
selenium.start({
  seleniumArgs: ['-cp /custom/lib1.jar']
}, function(err, child) {
});
```

Let me know if that doesn't work.

existdissolve:
@stephenash Thanks for the response! I tried both of those approaches, but didn't have any success.  I believe you have to use the _-cp /my/classpaths: org.openqa.grid.selenium.GridLauncher_ to get it to load the custom servlet successfully. When I follow the examples on the Selenium Grid page that discusses this, it works as expected. Thanks!

stephenash:
Closing since this appears to be supported as long as the syntax (or similar) listed in https://github.com/vvo/selenium-standalone/issues/202#issuecomment-238744407 is used.

PR #203, opened 2016-08-10
#202: Add support for classPath and update readme

joelwatson:
Suggested update for: https://github.com/vvo/selenium-standalone/issues/202

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,8 @@ as `selenium.install`.
 
 By default all drivers are loaded, you only control and change the versions or archs.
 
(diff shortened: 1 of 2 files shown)


vvo:
Hi @joelwatson thanks for that!

I am wondering though: given the comment here https://github.com/vvo/selenium-standalone/issues/202#issuecomment-238744407 it seems that the user made it work by using the `-- -cp..`.

Is your PR solving something different? Can you give me more context? Thanks

joelwatson:
Out of the box, I don't believe you can start it successfully with servets (or custom proxies) by providing *both* the -jar and the -cp; I believe it has to be either/or.

vvo:
@joelwatson Can you try it somehow before we include this code, just in case we might be adding some code that is not needed.

I just want to be sure/have proofs that we need this code.

vvo:
reopen if more comments

PR #210, opened 2016-08-24
Update driver and selenium versions in code examples

leventebalogh:
Update the examples so they contain the latest working versions of drivers and Selenium.


vvo:
THX

PR #211, opened 2016-08-24
Mention default config in the Readme

leventebalogh:
The file which is actually holding the up-to-date default config is not mentioned anywhere in the Readme. (It is `lib/default-config.js`)


vvo:
THX

PR #221, opened 2016-09-14
Lastest version of Chrome driver

Goye:
Updated default value for Chrome driver version

### DIFF
--- a/lib/default-config.js
+++ b/lib/default-config.js
@@ -3,7 +3,7 @@ module.exports = {
   version: '2.53.1',
   drivers: {
     chrome: {
-      version: '2.23',
+      version: '2.24',
       arch: process.arch,
       baseURL: 'https://chromedriver.storage.googleapis.com'
     },


vvo:
Will this work? Since it also need to handle mac64 download no? has you said

Goye:
k, let me see how to handle it :)

Goye:
@vvo sorry I was checking the issue and it was a problem with my version.

because in the 5.5.0 - computer-download-urls - line 25 you don't have a validation for that but in the 5.6.3, same file - line 28 you added: 

`if (opts.drivers.chrome.version < 2.23) {
     mac32 = true;
 }`  

so it's fixed right now

stephenash:
@Goye Sounds like everything is good then. If that is correct, can this PR and https://github.com/vvo/selenium-standalone/issues/220 be closed?

Goye:
@stephenash Yeah but with this PR I just updated the default version to the latest

vvo:
Not sure to follow, do we need to merge this to still default to 2.34?

Goye:
@vvo yep

PR #223, opened 2016-09-15
Test cases for constructing download URLs

stephenash:
There has been a number of PRs lately that improved how the URLs for binaries are generated. These helped us support older/newer versions of drivers where the structure of the filenames have been changed.

Since there is likely more of these types of changes necessary in the future, it is a good time to have some test cases in place to ensure that any future improvements support existing workflows as much as possible. This PR adds a non-exhaustive number of tests based on the conditionals that I found in the code.

NOTE: These tests do not verify that the module is returning valid URLs that will respond with the desired binary files. It is just testing that the URL is constructed how we expected.

### DIFF
--- a/lib/compute-download-urls.js
+++ b/lib/compute-download-urls.js
@@ -11,6 +11,30 @@ var urls = {
 
 var mac32;
 
+/**
+ * Computes the URL to download selenium.jar and drivers.
+ * 
+ * When passing in the `opts.drivers` param, it must be of the following format. Each of the drivers listed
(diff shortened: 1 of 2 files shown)


stephenash:
@jucrouzet In https://github.com/vvo/selenium-standalone/pull/222 you added a lot of logic regarding how the URLs for geckodriver were constructed. Can you review those test cases to ensure that I am covering everything that you added?

vvo:
awesome.. !

Issue #227, opened 2016-09-20
expectedRequests is incorrect when less than 3 opts.drivers is provided


johnelm:
In install.js, expectedRequests is incorrect when less than 3 opts.drivers is provided via the API.

Prevents the progress bar from ever showing, due to the if statement on line 329:

```javascript
  function updateProgressPercentage(chunk) {
    if (expectedRequests === startedRequests) {
      opts.progressCb(total, progress, chunk);
    }
  }
```

stephenash:
Closed via #229

PR #228, opened 2016-09-20
correct the expectedRequests counter when less than 3 opts.drivers provided

johnelm:
suggested update for: #227


johnelm:
I found an issue with this solution, related to the exclusion of ie driver on non-windows platforms - closing the PR.  Will submit again shortly.

PR #229, opened 2016-09-21
correct expectedRequests when less than 3 opts.drivers provided

johnelm:


johnelm:
suggested fix for #227

lib/install.js
```diff
@@ -46,6 +46,11 @@ function install(opts, cb) {
     opts.drivers = defaultConfig.drivers;
   }
 
+  if (process.platform !== 'win32') {
+    delete opts.drivers.ie;
```
vvo: Hi! What's the difference between delete opts.drivers.ie and deleting fsPaths/urls. I am unsure about the consequences between the two. Since you worked on this, do you have more details?

Thanks, this looks nice

johnelm: hey there!

yeah, in my first PR (which I cancelled) I was still deleting the ie member from the fsPaths and url objects separately.  

But since their ie members are created based on the opts.drivers object (i.e., their ie members are only created if there's an ie member in the opts.drivers object), deleting ie from opts.drivers has the same effect..  And doing this way seems preferred, because the expectedRequests counter was being decremented on non-windows platforms, regardless of whether there was an ie driver in the opts object.  I think this is the cause of #227 .

So the essence of the suggested fix is to base expectedRequests on how many drivers are present in opts, via Object.keys (plus 1, since we download the selenium-server-standalone jar regardless).

Cheers!

vvo: Thanks!

johnelm: no problem, thanks for your speedy response!  I'm pointing at your latest again in my work project.  👍 

Cheers!

Issue #230, opened 2016-09-23
Specify config file


kirkstrobeck:
What is the flag to specify the path to config for install?

stephenash:
Are you looking for something like `selenium-standalone install --config=...`? That doesn't exist.

You could use [the API](https://github.com/vvo/selenium-standalone#application-programming-interface-api) instead to do something like

```javascript
var selenium = require('selenium-standalone');
var opts = require('../path/to/my-custom-config.js');

selenium.install(opts, cb);
```

jucrouzet:
The flag is now supported : 
```
selenium-standalone install --config=/path/to/config.json
selenium-standalone start --config=./config/seleniumConfig.js
```

PR #236, opened 2016-10-21
Add a config file option (#230)

jucrouzet:
Add a `--config` option in CLI to specify options in a config file (See #230)

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,16 +46,8 @@ selenium-standalone install --config=./config/seleniumConfig.js
 
 ```
 
-Config file can be a JSON file or a [module file](https://nodejs.org/api/modules.html#modules_file_modules) that exports options as an object.
+Config file can be a JSON file or a [module file](https://nodejs.org/api/modules.html#modules_file_modules) that exports options as an object :
 
-JSON example :
-```json
-{
-  basePath: "/tmp/"
-}
-```
-
-JS File module example :
 ```js
 module.exports = {
   drivers: {
@@ -67,8 +59,6 @@ module.exports = {
   },
 }
 ```
-
-If your config file includes dynamic values (like `process.arch`) or functions (eg `spawnCb: function() {...`), you must use the JS File module format.
 
 ## Application Programming Interface (API)
 
--- a/README.md
+++ b/README.md
@@ -40,8 +40,35 @@ selenium-standalone start -- -role hub
 selenium-standalone start -- -role node -hub http://localhost:4444/grid/register
 selenium-standalone start -- -role node -hub http://localhost:4444/grid/register -port 5556
 
+# specify options in config file
+selenium-standalone install --config=/path/to/config.json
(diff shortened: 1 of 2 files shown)


jucrouzet:
IMHO, seemed a good options, especially when automating tests in Gulp/Grunt/etc. where we can change a very long command to a simple one.

bin/selenium-standalone
```diff
@@ -27,17 +28,33 @@ which('java', function javaFound(err, javaPath) {
...
+        throw new Error('Config file does not exports an object');
+      }
+    } catch(err) {
+      console.log('Error parsing config file :', err);
+      return;
+    }
+  }
+
+  // Merge default options, command line options and options from config file
+  options = merge({}, defaultConfig, options, configFromFile);
```
stephenash: `options` and `configFromFile` seem backwards to me. I would expect that CLI args have precedence over everything else.

README.md
```diff
@@ -40,9 +40,36 @@ selenium-standalone start -- -role hub
...
+selenium-standalone install --config=./config/seleniumConfig.js
 
 ```
 
+Config file can be a JSON file or a [module file](https://nodejs.org/api/modules.html#modules_file_modules) that exports options as an object.
+
+JSON example :
+```json
+{
+  basePath: "/tmp/"
```
stephenash: `basePath` should be in quotes to be valid JSON

stephenash:
Had a few small comments about the code, but overall looks like a good direction. Can you please add some unit tests as well, especially to validate that options are being overridden as expected?

vvo:
Nice! What's your take on JSON versus JavaScript configuration files? I do prefer JavaScript configuration files, most of the time it's more flexible. So I wonder: should we support JSON and JavaScript or just JavaScript and wait for someone to complain with a usecase?

bin/selenium-standalone
jucrouzet: That's right, will reverse the arguments.

README.md
jucrouzet: Yes... Thank you, will correct it.

jucrouzet:
@vvo : I personnaly prefer JS file also, but as `require()` accepts both, thought it would be better to leave it untouched. We can just use a JS example in the README

vvo:
> I personnaly prefer JS file also, but as require() accepts both, thought it would be better to leave it untouched. We can just use a JS example in the README

Works for me

jucrouzet:
Changes done.

vvo:
LGTM

vvo:
One of you guys want to do the release? That would help!

vvo:
Sorry @jucrouzet can you reopen your PR and add some tests as suggested? thx

jucrouzet:
@vvo Yup, I saw it. I'm in a rush at work right now, but will do it probably tomorrow.

vvo:
Ok just wanted to make sure you were not too much pissed-off by how it went. No need to rush

PR #241, opened 2016-10-26
Update geckodriver to v0.11.1

azakus:


vvo:
This is safe I believe? No change in binaries etc? :)

awilkins:
It's safe, and even required - geckodriver 0.10 just flat out doesn't work with Selenium 2.5.3 or 3.0.1 because they pass the --webdriver-port option to it - which only arrives as a hidden option in 0.11.1 ; version 0.10 will just fail to start with 

```
error: Found argument '--webdriver-port' which wasn't expected, or isn't valid in this context

USAGE:
    0.10.0-x64-geckodriver [FLAGS] [OPTIONS]

For more information try --help
error: Found argument '--webdriver-port' which wasn't expected, or isn't valid in this context

USAGE:
    0.10.0-x64-geckodriver [FLAGS] [OPTIONS]

For more information try --help
```

On the other hand ... Gecko is just flat out broken anyway. It doesn't support the Actions class, which means many tests (e.g. ones that send keypresses) are broken anyway.

If you have to test on Firefox AFAIK the only current way to get full service testing is going back to 47.0.1 (47.0 has a broken WebDriver...)

PR #245, opened 2016-11-16
adding copyright holders

julka:
Would you consider adding someone or some organization directly into the license file as the copyright holder?

This PR is just a suggestion, adding those listed in the authors file as copyright holders.

Thanks.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,5 @@
+Copyright (c) 2016 Vincent Voyer <vincent@zeroload.net> and Josh Chisholm <joshuachisholm@gmail.com>
+
 This software is released under the MIT license:
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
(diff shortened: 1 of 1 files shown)


vvo:
Hi @julka thanks for that! What is the effect of adding this legally or what does it provides for you?

julka:
Hi @vvo. 

First off, I'm not a lawyer. 

The correct form for an MIT license includes an attribution. https://opensource.org/licenses/MIT

My assumption of what it means to leave off the attribution is one of two extremes:

1) Whoever contributed a line in the codebase holds the copyright (and might be able to change the licensing for that line at any time they wish)
or 
2) The license is meaningless and might as well not be there at all.

My understanding of the legal impact is that this change would limit the copyright holder to the two of you instead of implicitly to whoever committed the code. Some projects explicitly attribute copyright to a specific person or org and then also contributors or authors. Here are several examples:

https://github.com/twbs/bootstrap/blob/v4-dev/LICENSE
https://github.com/jsdoc3/jsdoc/blob/master/LICENSE.md
https://github.com/jquery/jquery/blob/master/LICENSE.txt
https://github.com/dtabuenc/karma-html-reporter/blob/master/LICENSE
https://github.com/mochajs/mocha/blob/master/LICENSE

I'd be happy to update this PR to include something like that if you'd like.

This change helps me deal with my employer's legal/security approval BS so I can use this as a dev dependency in a project.

vvo:
> This change helps me deal with my employer's legal/security approval BS

:pray:

vvo:
THX

julka:
Thank you!

PR #251, opened 2016-12-22
Fix download urls

### LINKED ISSUE #250: platform is set to 'win64' if process.platform === 'win32' while downloading geckodriver
Could anyone explain the piece of code in https://github.com/vvo/selenium-standalone/blob/master/lib/compute-download-urls.js#L159

> } else if(process.platform === 'win32') {
>     platform = 'win64';
>     type = '.zip';
>   }

why do we set platform = 'win64' if process.platform === 'win32'?

jucrouzet:
Several fixes in download urls (and a little bit of refactoring).

It should fix #250

### DIFF
--- a/test/compute-download-urls-test.js
+++ b/test/compute-download-urls-test.js
@@ -340,7 +340,7 @@ describe('compute-download-urls', function() {
         assert(actual.firefox.indexOf('macos') > 0);
       });
 
-      it('uses `osx` platform for versions >= 0.6.0', function() {
+      it('uses `osx` platform for versions <= 0.6.0', function() {
         opts.drivers.firefox = {
           baseURL: 'https://localhost',
           version: '0.5.0',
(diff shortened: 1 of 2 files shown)


test/compute-download-urls-test.js
```diff
@@ -288,6 +339,17 @@ describe('compute-download-urls', function() {
         var actual = computeDownloadUrls(opts);
         assert(actual.firefox.indexOf('macos') > 0);
       });
+
+      it('uses `osx` platform for versions >= 0.6.0', function() {
```
stephenash: Test name should be `<` or `<=` to match the params?

jucrouzet: Right, nice spot on this typo. Fixed

jucrouzet:
:tada: !

Issue #238, opened 2016-10-25
Unit tests for new --config option


stephenash:
I [had requested](https://github.com/vvo/selenium-standalone/pull/236#pullrequestreview-5308663) new unit tests for your work with the new `--config` option. The PR got merged in before that was done.

@jucrouzet, can you take a look at adding those so that we don't accidentally break your changes in the future? It looks like you might have to extract the config parsing code into a separate module in order to make it testable in isolation.

Thanks!

vvo:
Sorry I was too quick to check that, just reverted the commit for now so that we can start over. Sorry about it

PR #252, opened 2016-12-27
Support for --config option

jucrouzet:
Support for the `--config=<path>` option for specifying a config file instead of command line arguments (#230) 
Included unit tests (#238) requested by @stephenash

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,8 @@ selenium-standalone start -- -role hub
 selenium-standalone start -- -role node -hub http://localhost:4444/grid/register
 selenium-standalone start -- -role node -hub http://localhost:4444/grid/register -port 5556
 
-# specify options in config file
+# If you have a complex configuration with numerous options or if you want to keep a clear configuration changes history,
+# you can specify the options in a configuration file :
 selenium-standalone install --config=/path/to/config.json
 selenium-standalone start --config=./config/seleniumConfig.js
 
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ selenium-standalone start --config=./config/seleniumConfig.js
 
 ```
 
-Config file can be a JSON file or a [module file](https://nodejs.org/api/modules.html#modules_file_modules) that exports options as an object :
+Config file can be a JSON file or a [module file](https://nodejs.org/api/modules.html#modules_file_modules) that exports options as an object:
 
 ```js
 module.exports = {
--- a/README.md
+++ b/README.md
@@ -40,7 +40,24 @@ selenium-standalone start -- -role hub
 selenium-standalone start -- -role node -hub http://localhost:4444/grid/register
 selenium-standalone start -- -role node -hub http://localhost:4444/grid/register -port 5556
 
+# specify options in config file
+selenium-standalone install --config=/path/to/config.json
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -40,9 +40,26 @@ selenium-standalone start -- -role hub
 selenium-standalone start -- -role node -hub http://localhost:4444/grid/register
 selenium-standalone start -- -role node -hub http://localhost:4444/grid/register -port 5556
 
+# specify options in config file
+selenium-standalone install --config=/path/to/config.json
+selenium-standalone install --config=./config/seleniumConfig.js
```
stephenash: Can you change this line to `selenium-standalone start ...` just to make it more explicit that the config file is supported in both actions.

bin/selenium-standalone
```diff
@@ -3,41 +3,74 @@
...
 
-  // Merge in default options
-  options = merge({}, defaultConfig, options);
+  // If a config file was specified, load it
+  var configFromFile = {};
+
+  if (argv.config) {
+    try {
+      configFromFile = require(
+        isAbsolute(argv.config) ? argv.config : path.resolve(process.cwd(), argv.config)
```
jucrouzet: It's only a workaround for the UNC bug on paths on windows (`\\host\path`)
```
C:\Users\jcrouzet>node -v
v6.9.2

C:\Users\jcrouzet>node
>  path.resolve(process.cwd(), 'c:\\site\\config')
'c:\\site\\config'
> path.resolve(process.cwd(), '\\\\nfs-share\\config')
'C:\\nfs-share\\config'
>
```

I agree it's a very rare case (just been traumatized by this one in past, sort of pavlov reflex).
I can remove it if it seems over-engineered

README.md
jucrouzet: Yep, done

bin/selenium-standalone
stephenash: Huh! That interesting. I found https://github.com/nodejs/node-v0.x-archive/issues/25552 (very old version of node) that mentions that the UNC path requires the folder to be valid.

Oddly, when I try your example I get back the expected result. Nothing in the release notes between your `node@6.9.2` and my `node@6.9.4` would indicate this was a bug fixed recently.

```text
> path.resolve(process.cwd(), '\\\\nfs-share')
'C:\\nfs-share'
> path.resolve(process.cwd(), '\\\\nfs-share\\config')
'\\\\nfs-share\\config\\'
```

jucrouzet: I'm on Win7, maybe ...
Anyway, it creates more confusion that it may solve cases ; I will remove it.

jucrouzet:
Just did some extended tests @ work (win10 / win8) ; the bug is only seen on Win7 with UNC paths without a valid hostname... So we can keep the "only path.resolve()" version on the current commit.
If it's ok for you guys, the PR is ready.

jucrouzet:
@vvo : Ok to release it also ?

vvo:
Yep seems good! Can we maybe add a small note about: given the project type, when is it useful to use a config file? In which situations if you have one in mind.

jucrouzet:
Done !
Let's go for a release

PR #256, opened 2017-01-13
Bump request and mocha to avoid npm warnings

Marsup:
Hi,

Just a quick bump of a few dependencies that were causing npm warnings on recent node versions.
Had to add a little code for request since its [version 2.59.0](https://github.com/request/request/blob/master/CHANGELOG.md#v2590-20150720) fixed HEAD behavior, which is actually rejected by S3, so a GET request is needed.
I have very little experience with this module so maybe my indexOf strategy is not the best, but it works for the default case.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -98,16 +98,16 @@ function install(opts, cb) {
   });
 
   function onlyInstallMissingFiles(opts, cb) {
-    async.series([
+    async.waterfall([
       checksum.bind(null, opts.to),
-      etag.bind(null, opts.from, requestOpts)
-    ], function (error, results) {
+      isUpToDate.bind(null, opts.from, requestOpts)
+    ], function (error, isLatest) {
       if (error) {
         return cb(error);
       }
 
       // File already exists. Prevent download/installation.
-      if (results[0] === results[1]) {
+      if (isLatest) {
         logger('---');
         logger('File from ' + opts.from + ' has already been downloaded');
         expectedRequests -= 1;
@@ -391,7 +391,7 @@ function logInstallSummary(logger, paths, urls) {
 
 function checksum (filepath, cb) {
   if (!fs.existsSync(filepath)) {
-    return cb();
+    return cb(null, null);
   }
 
   var hash = crypto.createHash('md5');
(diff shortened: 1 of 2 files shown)


lib/install.js
```diff
@@ -415,7 +415,18 @@ function unquote (str, quoteChar) {
 }
 
 function etag (url, requestOpts, cb) {
-  request.head(url, requestOpts).on('response', function (res) {
+  var verb = 'head';
+
+  if (url.indexOf('https://github.com') >= 0) {
```
jucrouzet: Indeed, the url.indexOf is not the best solution, you can use the Url module : 
```js
var Url = require('url');

var parsedUrl = Url.parse(url);

if (parsedUrl.host === 'github.com') {

}
```

BTW, why only on github.com ?

Marsup: I only found github to be redirecting to S3, the others are fine.

Marsup: Would you be fine with inlined require as well ?

jucrouzet:
@Marsup : Can you update package.json (version was bumped on last release) and add unit tests to be sure we cover this ?

@vvo / @stephenash : What's your POV on this ?

Marsup:
Rebased. I considered it already tested, as only upgrading request breaks the test, and this code fixes it, wdyt ?

stephenash:
@Marsup Thanks for the PR and your work here!

I'm not a big fan of having a workarounds for a list of specific servers. It is unfortunate that their S3 bucket is not set up to support `HEAD` requests, but I wouldn't want to be adding even more code branches to the code when the chromedriver servers act strange in the future.

One option would be to use a `GET` request for everything up front, but add the [`If-None-Match` header](https://en.wikipedia.org/wiki/HTTP_ETag#Typical_usage) for checking if the ETags match or not. If the server gives back a `200` response, then we save the output as before. A `304` means the local file matches and we can move along. Anything else is likely an error and should be thrown.

```bash
$ curl -qvL --header "If-None-Match: \"63a0b96eab18f8420b9bba2f0f5d380c\"" https://selenium-release.storage.googleapis.com/2.53/selenium-server-standalone-2.53.1.jar
*   Trying 172.217.3.176...
* Connected to selenium-release.storage.googleapis.com (172.217.3.176) port 443 (#0)
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA
* Server certificate: *.storage.googleapis.com
* Server certificate: Google Internet Authority G2
* Server certificate: GeoTrust Global CA
> GET /2.53/selenium-server-standalone-2.53.1.jar HTTP/1.1
> Host: selenium-release.storage.googleapis.com
> User-Agent: curl/7.43.0
> Accept: */*
> If-None-Match: "63a0b96eab18f8420b9bba2f0f5d380c"
>
< HTTP/1.1 304 Not Modified
< X-GUploader-UploadID: AEnB2UoxtSY1UZOBerjUEkMgRPGT-PqLrYn9tP9rjqq1t79vp-cVe7RzgJt3f4Pi4OmnTTG3Sg3u9yWCVvCAGYZ1znUh451hoUJG6RWF4vJo7DN_pMcmsN8
< Expires: Wed, 18 Jan 2017 20:47:34 GMT
< Date: Wed, 18 Jan 2017 19:47:34 GMT
< Cache-Control: public, max-age=3600
< Last-Modified: Thu, 30 Jun 2016 19:28:58 GMT
< Content-Length: 0
< Server: UploadServer
< Alt-Svc: quic=":443"; ma=2592000; v="35,34"
<
* Connection #0 to host selenium-release.storage.googleapis.com left intact
```

That type of refactoring is likely much more than you wanted to bite off, which is totally cool. We can merge this as is to get a better experience for newer NodeJS versions and then circle back with a more generalized approach at a later point.

Marsup:
@stephenash is that closer to what you have in mind ?

stephenash:
@Marsup That looks great; thanks for extra refactoring! I'll merge and get this released.

PR #259, opened 2017-01-22
Adding Yarn, updating dependencies, using 3.0.1 selenium by default

KennyRules:
* Adding [yarnpkg.com](yarn) support via auto-generated lockfile. 
* Ran `yarn upgrade-interactive` to update the following dependencies:
    * async
    * commander
    * lodash
    * minimist
    * mkdirp
    * progress
    * request
    * tar-stream
    * urijs
    * which

* Using selenium 3.0.1 for standalone-server and IE driver.
* Updated README to use newer 3.0.1 selenium and 2.27 chrome driver in examples.
* Modified travis file to use java environment to use jdk 8, which is needed for selenium 3+ it seems.


vvo:
Seems like good chore service here! Thanks a lot. If a specific new java version is required by selenium 3.0.0, will it throw any meaningful error to users? I just want to ensure that it throws the same error as when used without selenium-standalone npm package, so that people can just search on google and find the culprit.

KennyRules:
Thank you for the functionality!

Unfortunately I was not able to get any useful information when running Selenium 3 with Java 7.

At the very least I updated the PR to add the Java requirement to the README as well as adding some details to the error message.

README.md
```diff
@@ -185,8 +196,24 @@ After installing selenium-standalone globally, execute the following commands to
...
+
+### Ensure you have the minimum required Java version
+
+With the release of Selenium 3+, the minimum required version of Java is 8, as [as 7 has ceased public updates](https://java.com/en/download/faq/java_7.xml).
+
+If an older selenium version is needed, you can check the requirements on the [official Selenium changelog](https://raw.githubusercontent.com/SeleniumHQ/selenium/master/java/CHANGELOG).
+
+Here is a reference sheet for the more recent Selenium version:
+
+| Selenium version | Minimum Java Required |
```
vvo: Awesome!

README.md
```diff
@@ -172,11 +179,15 @@ pkill -f selenium-standalone
...
 ## Available browsers
 
+---
+
 By default, google chrome, firefox and phantomjs are available
 when installed on the host system.
 
 ## Tips
 
+---
```
vvo: Seems like you have some leftovers of auto formatting?

KennyRules: Oh! That was semi-intentionally me - The separator made the sections pop-out a bit more, though that might just be me.

Totally okay with reverting that change if desired!

KennyRules:
Ah - Yeah, my markdown previewer isn't as nice as what actually shows up on GitHub! Will revert.

vvo:
@stephenash @jucrouzet any comments? Should this be a new major version?

stephenash:
LGTM. I would say it is a major version update despite our API not changing; enough people could be unexpectedly hit by the new java 8 requirements.

This PR will also close https://github.com/vvo/selenium-standalone/issues/246.

jucrouzet:
LGTM also, but definitively a major version.
BTW, if you set a major bump, maybe we should set an `engine` property to >= 4  so that we can switch to es6 ?

vvo:
I am fine with:
- bumping major
- release new version

Ultimately we can bump engine >= 4 but in the end it's not even needed if you publish the transpiled version instead of the source files. This lets you write the program in any way you want, as long as you publish to ES5.

But still, if you are up to it and don't want to have to transpile but just publish and if you ensure it's working with automated tests we are good.

I let you handle the merge then!

KennyRules:
@vvo : Just want to make sure, you / others are not waiting on my end for a engine, or any other, change to this PR, correct?

Didn't want to seem to disappear if you were in fact waiting, as I had nothing on my list of things to do!

.travis.yml
```diff
@@ -1,4 +1,16 @@
 sudo: false
 
-language: node_js
-node_js: stable
+language: java
```
jucrouzet: Changing language is not necessary, could you just keep the `jdk` property in addition ?

README.md
```diff
@@ -185,8 +186,22 @@ After installing selenium-standalone globally, execute the following commands to
 ln -s /usr/local/bin/selenium-standalone /etc/init.d/
 update-rc.d selenium-standalone defaults
 ```
+
 For more information: https://stackoverflow.com/questions/3666794/selenium-server-on-startup/30392437#30392437
 
+### Ensure you have the minimum required Java version
+
+With the release of Selenium 3+, the minimum required version of Java is 8, as [as 7 has ceased public updates](https://java.com/en/download/faq/java_7.xml).
```
jucrouzet: Little typo here  `as as`

jucrouzet:
Sorry, didn't pay attention to the .travis.yml before.

jucrouzet:
@vvo : I'm not  a big fan of transpiled js, but right, the issues show that there is still some users with a node 0.1* version so yes, let's not touch the target and transpile if we switch to es6.

.travis.yml
KennyRules: Unfortunately this did not seem to work. Investigation led to this travis issue:
https://github.com/travis-ci/travis-ci/issues/3827

An alternative could be using node_js language and then manually hooking up java.

jucrouzet: I see. 
The problem with this YAML is that it breaks the possibility to test multiple node version.
Can you try if it works with this version : 
```yml
sudo: false

language: java

jdk:
  - oraclejdk8

node_js:
  - stable
  - 5
  - 4
  - 0.12

before_install:
  - nvm install $TRAVIS_NODE_VERSION

install: "npm install"

script: "npm test"
```

jucrouzet: (BTW it's normal if it fails on 0.12 => #460 )

jucrouzet:
Green flag, let's go for a release !
Thanks for the PR !

Issue #261, opened 2017-01-30
Wrong IE path or missing file in v6.0.0


jlchereau:
https://selenium-release.storage.googleapis.com/3.0/IEDriverServer_x64_3.0.1.zip does not seem to be available.

```
D:\WebStorm\Project>selenium-standalone install
----------
selenium-standalone installation starting
----------

---
selenium install:
from: https://selenium-release.storage.googleapis.com/3.0/selenium-server-standalone-3.0.1.jar
to: C:\Users\xxxxx\AppData\Roaming\npm\node_modules\selenium-standalone\.selenium\selenium-server\3.0.1-server.jar
---
chrome install:
from: https://chromedriver.storage.googleapis.com/2.27/chromedriver_win32.zip
to: C:\Users\xxxxx\AppData\Roaming\npm\node_modules\selenium-standalone\.selenium\chromedriver\2.27-x64-chromedriver
---
ie install:
from: https://selenium-release.storage.googleapis.com/3.0/IEDriverServer_x64_3.0.1.zip
to: C:\Users\xxxxx\AppData\Roaming\npm\node_modules\selenium-standalone\.selenium\iedriver\3.0.1-x64-IEDriverServer.exe
---
firefox install:
from: https://github.com/mozilla/geckodriver/releases/download/v0.13.0/geckodriver-v0.13.0-win64.zip
to: C:\Users\xxxxx\AppData\Roaming\npm\node_modules\selenium-standalone\.selenium\geckodriver\0.13.0-x64-geckodriver
C:\Users\xxxxx\AppData\Roaming\npm\node_modules\selenium-standalone\bin\selenium-standalone:105
        throw err;
        ^

Error: Could not download https://selenium-release.storage.googleapis.com/3.0/IEDriverServer_x64_3.0.1.zip
    at Request.<anonymous> (C:\Users\xxxxx\AppData\Roaming\npm\node_modules\selenium-standalone\lib\install.js:273:21)
    at emitOne (events.js:96:13)
    at Request.emit (events.js:188:7)
    at Request.onRequestResponse (C:\Users\xxxxx\AppData\Roaming\npm\node_modules\selenium-standalone\node_modules\request\request.js:986:10)
    at emitOne (events.js:96:13)
    at ClientRequest.emit (events.js:188:7)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:474:21)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:99:23)
[...]

PR #262, opened 2017-01-30
Update default value for IE driver to one that exists.

stephenash:
Provides fix for #261.

Also adds some new test cases that try to download all the binaries that come with the defaults. Hopefully the new tests will help us catch any problems when the default versions change or the paths to the remote files change.
* Selenium Server
* IE (win32, win64)
* Chrome (linux32, linux64, mac, win32)
* Firefox (linux32, linux64, mac, win32, win64, arm)


jucrouzet:
:+1:

vvo:
LGTM

Issue #253, opened 2017-01-06
Does not set permission properly for downloaded drivers


CTaylor1118:
When downloading drivers, it looks like the code is supposed to try to modify the permissions so the files are executable but it is not. 

Version: 5.8.0
Node Version: 0.12.17

stephenash:
Can you share more info about your environment (platform, how you or another tool is calling `selenium-standalone`, etc.) and what types of errors you are seeing?

I tried with my local machine (OSX 10.10) and didn't immediately notice anything not working.

```bash
$ node -v
v0.12.17

$ cat node_modules/selenium-standalone/package.json| grep version
  "version": "5.8.0",

$ ./node_modules/selenium-standalone/bin/selenium-standalone install
----------
selenium-standalone installation starting
----------

---
selenium install:
from: https://selenium-release.storage.googleapis.com/2.53/selenium-server-standalone-2.53.1.jar
to: /private/tmp/selenium-standalone-253/node_modules/selenium-standalone/.selenium/selenium-server/2.53.1-server.jar
---
chrome install:
from: https://chromedriver.storage.googleapis.com/2.25/chromedriver_mac64.zip
to: /private/tmp/selenium-standalone-253/node_modules/selenium-standalone/.selenium/chromedriver/2.25-x64-chromedriver
---
firefox install:
from: https://github.com/mozilla/geckodriver/releases/download/v0.10.0/geckodriver-v0.10.0-macos.tar.gz
to: /private/tmp/selenium-standalone-253/node_modules/selenium-standalone/.selenium/geckodriver/0.10.0-x64-geckodriver


selenium-standalone installation [==================  ] 88% 2.6s

-----
selenium-standalone installation finished
-----

$ ls -AlR node_modules/selenium-standalone/.selenium
total 0
drwxr-xr-x  4 sash  wheel  136 Jan  7 09:33 chromedriver
drwxr-xr-x  4 sash  wheel  136 Jan  7 09:33 geckodriver
drwxr-xr-x  3 sash  wheel  102 Jan  7 09:33 selenium-server

node_modules/selenium-standalone/.selenium/chromedriver:
total 29984
-rwxr-xr-x  1 sash  wheel  10782984 Jan  7 09:33 2.25-x64-chromedriver
-rw-r--r--  1 sash  wheel   4563462 Jan  7 09:33 2.25-x64-chromedriver.zip

node_modules/selenium-standalone/.selenium/geckodriver:
total 9184
[...]

PR #271, opened 2017-03-17
Ability to collect debug logs for bug reports

stephenash:
From time to time, issues are opened that are really hard to track down. I wanted to ability for users to quickly turn on debug logs that will show me exactly how `selenium-standalone` was being executed, along with some environment details.

Originally this was done to track down why #253 was happening, but in the future more types of logs can be enabled as needed.

_Example:_
```text
$ DEBUG=selenium-standalone:* selenium-standalone install --version=3.1.1
  selenium-standalone:env-details Platform: darwin 14.5.0 +0ms
  selenium-standalone:env-details Architecture: x64 +4ms
  selenium-standalone:env-details Node.js: v4.6.1 +2ms
  selenium-standalone:env-details CWD: /Users/sash/dev/git/selenium-standalone-testing +0ms
  selenium-standalone:env-details Module Path: /Users/sash/dev/git/selenium-standalone/lib +0ms
  selenium-standalone:env-details Package Version: 6.1.0 +1ms
  selenium-standalone:cli Started via CLI with:  [ '/Users/sash/.nvm/versions/node/v4.6.1/bin/node',
  '/Users/sash/.nvm/versions/node/v4.6.1/bin/selenium-standalone',
  'install',
  '--version=3.1.1' ] +203ms
  selenium-standalone:install Install API called with { baseURL: 'https://selenium-release.storage.googleapis.com',
  version: '3.1.1',
  drivers:
   { chrome:
      { version: '2.28',
        arch: 'x64',
        baseURL: 'https://chromedriver.storage.googleapis.com' },
     ie:
      { version: '3.0.0',
        arch: 'x64',
        baseURL: 'https://selenium-release.storage.googleapis.com' },
     firefox:
      { version: '0.13.0',
        arch: 'x64',
        baseURL: 'https://github.com/mozilla/geckodriver/releases/download' } },
  seleniumArgs: [],
  spawnOptions: { stdio: 'inherit' },
  logger: [Function: bound ],
  javaPath: '/Users/sash/.jenv/shims/java',
  progressCb: [Function: progressCb] } +12ms
----------
----------
...
```


vvo:
Since we already have a logger, this would more be a "debug logger" while the current logger is an API feature right?

stephenash:
> Since we already have a logger, this would more be a "debug logger" while the current logger is an API feature right?

Correct. `options.logger` seemed to be isolated to just the installation process. I wanted to be able to expose additional logs from other parts of the applications, potentially even before `options.logger` is defined.

I'm open to ideas on consolidating `options.logger` into broader logging improvements. That could make it easier to make improvements in the future if we don't need to thinks about `options.logger` vs `console.log` vs `debug`.

vvo:
I am fine separating both concerns, one logger is for normal conditions, second one is mostly for development debugging sessions. Ready to merge when you are.

PR #278, opened 2017-04-22
Update for lastest version of Selenium and Geckodriver

geskill:
Changelogs:
- [Selenium](https://raw.githubusercontent.com/SeleniumHQ/selenium/master/java/CHANGELOG)
- [Internet Explorer](https://raw.githubusercontent.com/SeleniumHQ/selenium/master/cpp/iedriverserver/CHANGELOG)
- [Geckodriver](https://github.com/mozilla/geckodriver/releases/tag/v0.16.0)

### DIFF
--- a/lib/default-config.js
+++ b/lib/default-config.js
@@ -1,6 +1,6 @@
 module.exports = {
   baseURL: 'https://selenium-release.storage.googleapis.com',
-  version: '3.3.1',
+  version: '3.4.0',
   drivers: {
     chrome: {
       version: '2.29',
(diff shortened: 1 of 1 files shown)


vvo:
👌 , no idea if this will break everything as usual, should we bump major?

geskill:
i wouldn't since 6.x comes with selenium 3.x already.

_remark:
This update fixes issues with Firefox, in order to work with FF 47+ (thats at least what i experienced). Before i had issues using any FF version in selenium 3. Therefore i used selenium 2 and FF 47._

geskill:
Currently, the Actions API is not completely supported by Geckodriver  https://github.com/mozilla/geckodriver/issues/159

vvo:
> Currently, the Actions API is not completely supported by Geckodriver mozilla/geckodriver#159

Does it means we need to do something different in term of versionning?

geskill:
No, because this update is an improvement, although there is this unimplemented feature. Before there is no conf (i know) that allow us to use FF with Selenium 3.x

PR #284, opened 2017-06-06
Add Microsoft Edge Web Driver

unkiwii:
Vincent, I'm using your library and wanted to test on Microsoft Edge. I added the possibility to install and use Microsoft Edge web driver following this: https://developer.microsoft.com/en-us/microsoft-edge/tools/webdriver/#downloads

Hope you found it usefull and merge it to the master branch.

I didn't added tests because other tests were failing for me before even touch a line of code.

### DIFF
--- a/lib/compute-download-urls.js
+++ b/lib/compute-download-urls.js
@@ -6,7 +6,8 @@ var urls = {
   selenium: '%s/%s/selenium-server-standalone-%s.jar',
   chrome: '%s/%s/chromedriver_%s.zip',
   ie: '%s/%s/IEDriverServer_%s_%s.zip',
-  firefox: '%s/%s/%s-%s-%s'
+  firefox: '%s/%s/%s-%s-%s',
+  edge: '%s/%s/MicrosoftWebDriver.%s'
 };
 
 var mac32;
@@ -75,6 +76,14 @@ function computeDownloadUrls(opts, askedOpts) {
       getFirefoxDriverName(opts.drivers.firefox.version),
       getFirefoxDriverVersionForUrl(opts.drivers.firefox.version),
       getFirefoxDriverArchitecture(opts.drivers.firefox.version, opts.drivers.firefox.arch)
+    );
+  }
+  if (opts.drivers.edge) {
+    downloadUrls.edge = util.format(
+      urls.edge,
+      opts.drivers.edge.baseURL,
+      getEdgeDriverDirectory(opts.drivers.edge.version),
+      getEdgeDriverExtension(opts.drivers.edge.version)
     );
   }
   return downloadUrls;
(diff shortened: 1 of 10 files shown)


lib/compute-download-urls.js
```diff
@@ -231,6 +240,36 @@ function getWindowsFirefoxDriverArchitecture(version, wantedArchitecture) {
...
+
+function getEdgeDriverExtension(version) {
+  var release = parseInt(version)
+  switch (release) {
+    case 15063:
+    case 14393:
+      return 'exe'
+    case 10586:
+    case 10240:
+      return 'msi'
```
stephenash: I see that these two versions are released as an `.msi` package, but there isn't any code to run that installer and get the path to the `MicrosoftWebDriver.exe` file that we really need.

You can look at how this post-download is handled for Firefox which is shipped as a `.zip` or `.tar.gz`. https://github.com/vvo/selenium-standalone/blob/baf462261ae86d401f369d1c76542fcb3b7be3af/lib/install.js#L186

lib/default-config.js
```diff
@@ -12,6 +12,11 @@ module.exports = {
       arch: process.arch,
       baseURL: 'https://selenium-release.storage.googleapis.com'
     },
+    edge: {
+      version: '14393',
```
stephenash: The default version should be the latest (`15063`).

lib/install.js
```diff
@@ -96,6 +97,10 @@ function install(opts, cb) {
     tasks.push(setDriverFilePermissions.bind(null, fsPaths.firefox.installPath));
   }
 
+  if (fsPaths.edge) {
+    tasks.push(setDriverFilePermissions.bind(null, fsPaths.edge.installPath));
+  }
```
stephenash: Is this block needed? There isn't an existing similar one for IE. Or is there a bug that the IE block is missing?

That `setDriverFilePermissions()` ends up calling chmod which I'm not sure what happens on Windows.

lib/compute-download-urls.js
```diff
@@ -231,6 +240,36 @@ function getWindowsFirefoxDriverArchitecture(version, wantedArchitecture) {
...
+  var release = parseInt(version)
+  switch (release) {
+    case 15063:
+      return '3/4/2/342316D7-EBE0-4F10-ABA2-AE8E0CDF36DD'
+    case 14393:
+      return '3/2/D/32D3E464-F2EF-490F-841B-05D53C848D15'
+    case 10586:
+      return 'C/0/7/C07EBF21-5305-4EC8-83B1-A6FCC8F93F45'
+    case 10240:
+      return '8/D/0/8D0D08CF-790D-4586-B726-C6469A9ED49C'
```
stephenash: Gross! This is going to be a headache to maintain if there is a lot of edge drivers released. Don't think there is really anything that can be changed here; just me wishing that Microsoft had the version number in the path instead of uuids.

unkiwii: I added a file at lib\microsoft-edge-releases.js that contains all the releases, that should be easy to mantain, just add an entry there when a new version is released and all should work

lib/compute-download-urls.js
unkiwii: Fixed this. For versions of `.msi` files the installer is run using `msiexec` and the filed installed is copied to the correct place.

lib/default-config.js
unkiwii: Changed to 15063

lib/install.js
unkiwii: You are right, that was removed

unkiwii:
All changes done and tests added.

I have tested it on a 64bit Windows 10 and all tests and runs worked perfectly.

stephenash:
@unkiwii Thanks again for the PR and the extra work to add test cases and the `.msi` installer logic. The changes look good; I'll get them merged in and released shortly.

vvo:
Thanks a lot @stephenash

PR #298, opened 2017-08-15
Update HISTORY.md

geskill:


stephenash:
I'll merge this since I happen to catch this on the same date that you have listed in the changelog file. In the future this should be unnecessary since it will be up to the person releasing the next version to ensure that the changelog is accurate.

PR #305, opened 2017-09-06
Add correct urandom devnode path when on linux

jkva:
When on Linux (in my case, Debian) Selenium won't start because it waits for entropy from `/dev/random` (which can block and depletes the entropy pool). Pointing it to `/dev/urandom` fixes this.

### DIFF
--- a/lib/start.js
+++ b/lib/start.js
@@ -77,6 +77,10 @@ function start(opts, cb) {
     args.push('-Dwebdriver.chrome.driver=' + fsPaths.chrome.installPath);
   }
 
+  if (process.platform === 'linux') {
+    args.push('-Djava.security.egd=file:///dev/urandom');
+  }
+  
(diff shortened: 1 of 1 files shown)


stephenash:
Though I appreciate the PR, I don't think this is the right choice for the `selenium-standalone` project to make.

For many cases (including your environment) this change may make sense, but that is not to say other environments would have problems due to this change and would prefer the existing default java/selenium choice.

Since we already expose the option to pass in these types of arguments that make your own environment work, I recommend you use that instead. See https://github.com/vvo/selenium-standalone#seleniumstartopts-cb

```javascript
seleniumStandalone.start({
   javaArgs: '-Djava.security.egd=file:///dev/urandom'
});
```

jkva:
@stephenash that's fair, and thanks for explaining your rationale.

PR #308, opened 2017-09-15
Fix selenium server status check URL when passing node config file

germanz:
As mentioned in selenium server documentation it is possible to pass the node configuration options via -nodeConfig parameter (https://github.com/SeleniumHQ/selenium/wiki/Grid2#configuring-the-nodes-by-json). 
Unfortunately selenium-standalone didn't had support for that parameter and was constructing the status check URL incorrectly. Therefore node was stopping after several failed attempts.

I changed selenium status page URL construction to take into account host, port parameters defined in node config file and added tests.

### DIFF
--- a/lib/get-selenium-status-url.js
+++ b/lib/get-selenium-status-url.js
@@ -1,26 +1,74 @@
 var URI = require('urijs');
+var fs = require('fs');
+
 var PROCESS_TYPES = exports.PROCESS_TYPES = {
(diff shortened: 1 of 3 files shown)


vvo:
Seems legit and tests are passing, wdyt @stephenash ?

jucrouzet:
Seems good for me (unit test and manual tests).

PR #309, opened 2017-09-22
fixes #304 - added timeout option to request

### LINKED ISSUE #304: install should timeout after x seconds
## Expected Behavior

I'm currently on a train and try to debug a selenium test. I run `install`, but because of some connection problem the `install` will run forever. It would be nice to have a default timeout (maybe after 10 seconds?), so a real error is thrown. It took me a while to figure out that the `install` runs forever and not some other part of my test.

## Current Behavior

Currently `install` runs forever on a (flaky?) connection.

## Steps to Reproduce

Hard to reproduce. But I can try ;)

- Visit Germany. 🍺 
- Go on a train trip from Munich to Leipzig. 🚋 
- Train should be an ICE. 🚅 
- Log into `WIFIonICE` to get free internet. 📶 
- (Oh, I'm also on corporate VPN for some extra fun. 💃 )
- Run `install`. 🎉 

😄

donaldpipowitch:
Not tested thoroughly (hard for me to reproduce, too), but it looks like I just need to configure `request`? Should fix #304

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,9 @@ selenium.install({
     }
   },
   proxy: 'http://localproxy.com', // see https://github.com/request/request#proxies
+  requestOpts: { // see https://github.com/request/request#requestoptions-callback
+    timeout: 10000
+  },
   logger: function(message) {
 
   },
@@ -149,6 +152,8 @@ arch [sometimes](https://code.google.com/p/selenium/issues/detail?id=5116#c9).
 
 `opts.logger` will be called if provided with some debugging information about the installation process.
 
+`opts.requestOpts` can be any valid [`request` options object](https://github.com/request/request#requestoptions-callback). You can use this for example to set a timeout.
+
 `cb(err)` called when install finished or errored.
 
 ### selenium.start([opts,] cb)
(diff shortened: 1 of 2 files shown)


jucrouzet:
Fixing an arbitrary timeout of 10s may be a blocking change for people under a slow connection, IMHO, it needs to be configurable.
For the unit tests, you can set a proxy (https://github.com/nodejitsu/node-http-proxy for example) with a delay on the response.

donaldpipowitch:
> it needs to be configurable.

Do you want to just make `timeout` configurable or maybe the whole request options? This _could_ also deprecate the `proxy` option which could be recreated when the whole request options would be configurable.

jucrouzet:
Maybe making the whole requestOptions configurable would be lighter and more flexible. @vvo / @stephenash ?

vvo:
Yep we should make it so that anyone can pass requestOptions via programmatic API

vvo:
And explain in README those edge cases

donaldpipowitch:
Updated 👍

vvo:
Thanks, perfect

PR #311, opened 2017-10-13
Add option to explicitly specify what drivers are installed.

robertmorgan:


robertmorgan:
Seems to have some stability issues in Travis that don't appear locally. I'll investigate and re-open later.

vvo:
@robertmorgan do you have to have another try?

PR #335, opened 2018-01-05
Update for lastest version of Chrome

paladox:


vvo:
tests are red here

paladox:
It seems to be because chromedriver is not available for Linux 32.

vvo:
> It seems to be because chromedriver is not available for Linux 32.

Will they provide it? Any information on this? I'd rather not delete some tests or feature here..

paladox:
@vvo not sure, i just see the 32 bit files for linux missing from 34 and 35.

paladox:
See https://bugs.chromium.org/p/chromedriver/issues/detail?id=2176

chromedriver stopped supporting linux32 due to chrome stopping support and due to low usage.

Maybe we will want to use an old driver for linux32 but not sure how to do that one?

stephenash:
@paladox If there are not going to be `chromedriver` releases for linux32 anymore, then we should be able to support old versions that have been released though.

Firefox did something similar in the past, so there is some code [0] that checks if a platform is known to work with the requested version. You could add some similar logic (and some tests) to verify that the old releases continue to work, but that the latest release does not support linux32.

[0] https://github.com/vvo/selenium-standalone/blob/master/lib/compute-download-urls.js#L175-L186

vvo:
If you still need this, open a new PR and fix the conflicts thanks!

PR #339, opened 2018-01-24
Update driver and selenium versions

plessbd:
update the selenium and webdriver versions to "current"
this obsoletes pull #334 as this also update chromedriver

Also, remove test for chromedriver ia32 as it is no longer being built by google

### DIFF
--- a/test/default-downloads-test.js
+++ b/test/default-downloads-test.js
@@ -129,21 +129,6 @@ describe('default-downloads', function() {
         });
       });
 
-      it('ia32 download exists', function(done) {
-        opts = merge(opts, {
-          drivers: {
-            chrome: {
-              arch: 'ia32'
-            }
-          }
-        });
-
-        computedUrls = computeDownloadUrls(opts);
-
-        assert(computedUrls.chrome.indexOf('linux32') > 0);
-        doesDownloadExist(computedUrls.chrome, done);
-      });
-
       it('x64 download exists', function(done) {
         opts = merge(opts, {
           drivers: {
--- a/lib/default-config.js
+++ b/lib/default-config.js
@@ -1,14 +1,14 @@
 module.exports = {
   baseURL: 'https://selenium-release.storage.googleapis.com',
-  version: '3.7.1',
+  version: '3.8.1',
   drivers: {
     chrome: {
-      version: '2.33',
+      version: '2.35',
       arch: process.arch,
       baseURL: 'https://chromedriver.storage.googleapis.com'
     },
     ie: {
-      version: '3.7.0',
+      version: '3.8.0',
       arch: process.arch,
       baseURL: 'https://selenium-release.storage.googleapis.com'
     },


test/default-downloads-test.js
```diff
@@ -129,21 +129,6 @@ describe('default-downloads', function() {
         });
       });
 
-      it('ia32 download exists', function(done) {
```
stephenash: Let's not remove a test just because it is failing. In https://github.com/vvo/selenium-standalone/pull/335 (another PR wanting to update the default versions), I indicated that if chrome is no longer shipping platform-specific version, then new logic and tests should be added in order to make sure that old versions can still be used in `selenium-standalone`.

plessbd:
Ahh, sorry didnt notice the other PR.  If I get some time I will see if I can add the extra logic to be able to have both.

How often do we have people installing old versions of the driver and/or browser though?

Generally, in my own experience, I have noticed most users end up on the newest version of the browser (all the browsers now seem to auto update) so I don't even get support requests for old browser versions (at least not in the past few years)

abjerstedt:
This probably should get resolved sooner than later... atm the moment this package locks the users into a chromedriver version that doesn't support the current or last chrome version.

stephenash:
> atm the moment this package locks the users into a chromedriver version that doesn't support the current or last chrome version.

@abjerstedt That is not true. The user still has the ability to overwrite the default version used in order to use specific versions that work in their own environment. The details on how to do that can be found in the [README](https://github.com/vvo/selenium-standalone#sample-configuration-object).

> Would you be fine if i threw up a PR that does nothing bug change the version numbers?

Just changing the default version number does not solve the problem where support for the `ia32` platform should be maintained for older versions of the driver. That is why the existing unit tests failed for the first commit in this PR.

abjerstedt:
@stephenash So the referenced example will persist through branch changes, and repeated dependency installations (in our case yarn install)? 

In our case we use selenium-standalone-service for webdriverio, and selenium-standalone is a nested dependency.

I've already seen my configurations overwritten multiple times in CI environment, and i would rather not write a setup script to manually alter this with every build invocation. (i actually switched to calling chromedriver so i could maintain direct control).

stephenash:
@abjerstedt Yes, those configuration examples live in your codebase, not `selenium-standalone`, so it should persist.

Since you are using webdriverio, you can use the `seleniumArgs` and `seleniumInstallArgs` in your `wdio.conf.js` file. See https://github.com/webdriverio/wdio-selenium-standalone-service#seleniumargs. Those match up to the objects referenced in this project's docs for `.start()` and `.install()`.

abjerstedt:
On a side note... is anyone else having trouble with chromedriver 2.36 or 2.37? My tests suddenly failing due to a connection errors.

vvo:
conflicts on this PR, let me know if you want to fix them and merge, or just open another PR

plessbd:
No longer needed #354 takes care of version bump.

PR #340, opened 2018-01-24
Remove ia32 linux chromedriver test

plessbd:
Remove ia32 linux chromedriver test as it is no longer supported by google

https://groups.google.com/forum/#!msg/chromedriver-users/JZUYvp_xr60/pyTRBxv1CwAJ


plessbd:
updated #339 since it is related to the chromedriver update

PR #363, opened 2018-05-25
Updated to the latest Geckodriver that works with Firefox 66.0.1

serbanghita:
Geckodriver 0.20.0 + Firefox 66.0.1 fails when attempting to click on an element which is visible in page but has `elem.getBoundingClientRect().height === 0`. I tested this with `wdio` on `document.body`

Geckodriver 0.20.1 + Firefox 66.0.1 works

Example of page in Firefox that reports height 0 on `document.body`:
screen shot 2018-05-25 at 11 37 57

Q: @vvo have you ever considered having some e2e tests of browser quirks in different Browser + Webdriver combinations?

### DIFF
--- a/lib/default-config.js
+++ b/lib/default-config.js
@@ -13,7 +13,7 @@ module.exports = {
       baseURL: 'https://selenium-release.storage.googleapis.com'
     },
     firefox: {
-      version: '0.20.0',
+      version: '0.20.1',
       arch: process.arch,
       baseURL: 'https://github.com/mozilla/geckodriver/releases/download'
     },


vvo:
> Q: @vvo have you ever considered having some e2e tests of browser quirks in different Browser + Webdriver combinations?

Hi there, no I never considered having that, since selenium-standalone is only a wrapper to ease the installation of selenium. Our scope is to only ensure it can be installed and launched, not that all the previous browser features from a version of selenium are still available. This is the responsibility of the selenium project itself.

I hope this is clear enough, thanks!

serbanghita:
@vvo yes clear, then I will pitch this to https://github.com/webdriverio/webdriverio (this is what I'm using) since it makes more sense to have them there

serbanghita:
@vvo can you release a minor version based on this? Thank you 🍺

vvo:
done

PR #364, opened 2018-05-30
recognise MicrosoftWebDriver 17134

dten:
allow edge webdriver 17134 to be recognised as valid version

### DIFF
--- a/lib/microsoft-edge-releases.js
+++ b/lib/microsoft-edge-releases.js
@@ -32,5 +32,10 @@ module.exports = {
   '16299': {
     url: 'https://download.microsoft.com/download/D/4/1/D417998A-58EE-4EFE-A7CC-39EF9E020768/MicrosoftWebDriver.exe',
     extension: 'exe'
+  },
+
+  '17134': {
+    url: 'https://download.microsoft.com/download/F/8/A/F8AF50AB-3C3A-4BC4-8773-DC27B32988DD/MicrosoftWebDriver.exe',
+    extension: 'exe'
   }
 };


amit1980in:
I tried your fix with web component tester and it didn't worked. I am still getting following error
Error: The following browsers were not found: edge. (All installed browsers found: chrome, ie)

I am trying this on windows 10 with OS version 17134

dten:
What did you try? Works for me when I use a custom config to install / start selenium standalone

amit1980in:
I am using web component tester to test my component written in polymer against edge browser. it works fine with edge browser 16.16299 but not with edge version 17.17134. My understanding is that web component internally use selenium to run test cases against specific web driver version. I tried running WCT command after I added edge version 17.17134 in microsoft-edge-releases.js file and I am getting below error while running wct command.

Error: The following browsers were not found: edge. (All installed browsers found: chrome, ie)

dten:
You'll want to change the default config also to make it use this version

https://github.com/vvo/selenium-standalone/blob/master/lib/default-config.js

Or use a custom config

amit1980in:
I changed default-config.js to point edge: {
      version: '17134'
    }
 but still I am getting below errror
Error: The following browsers were not found: edge. (All installed browsers found: chrome, ie)

amit1980in:
I would like to run rest against 17.17134 edge browser because this is what comes with windows 10 latest update. Please advise if selenium support it.

amit1980in:
Hi
Just wanted to inform you that I am able to launch edge browser 17.17134 after I restarted PC.

Thanks for your help.

Regards
Amit

PR #365, opened 2018-05-30
fix(install): prefer the user config rather that the default one

douglasduteil:
When using 

```sh
selenium-standalone install --version=3.6.0 --baseURL=http://my.mirror.com/selenium-server --drivers.chrome.version=2.33 --singleDriverInstall=chrome --drivers.chrome.baseURL=http://my.mirror.com/chromedriver`
```

Selenium ignore the `--drivers.chrome.baseURL` part and says :

```sh

----------
----------

---
selenium install:
from: http://my.mirror.com/selenium-server/3.6/selenium-server-standalone-3.6.0.jar
to: D:\x\y\z\node_modules\selenium-standalone\.selenium\selenium-server\3.6.0-server.jar
---
chrome install:
from: https://chromedriver.storage.googleapis.com/2.33/chromedriver_win32.zip
to: D:\x\y\z\node_modules\selenium-standalone\.selenium\chromedriver\2.33-x64-chromedriver

```

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -51,9 +51,10 @@ function install(opts, cb) {
   }
 
   if (opts.singleDriverInstall) {
-    if(defaultConfig.drivers[opts.singleDriverInstall]) {
+    var singleDriver = opts.drivers[opts.singleDriverInstall];
+    if(singleDriver) {
       opts.drivers = {};
-      opts.drivers[opts.singleDriverInstall] = defaultConfig.drivers[opts.singleDriverInstall];
+      opts.drivers[opts.singleDriverInstall] = singleDriver;
     }
   }
 


serbanghita:
👍 I'm looking into this, I need to add some tests for the install part so I don't mess the configurations.

douglasduteil:
Thanks @DAtink

I don't know..
It feels like using `selenium-standalone` behind a proxy is pretty rare...

vvo:
@douglasduteil If @DAtink if you need proxy usage you could also call selenium-standalone this way:

```sh
HTTP_PROXY=... selenium-standalone
```

But maybe you need a different option. Let me know

vvo:
Second PR talking about this, merging and will revert if internet screams

douglasduteil:
:tada: Thanks @vvo

vvo:
this was released in 6.17.0

PR #387, opened 2018-08-16
Fix non working Docker image

catalan-adobe:
**Issue Report:**

Testing environment:
* macOS: 10.12.6
* Docker: 18.06.0-ce-mac70 (26399) (stable channel)
* Clean clone of vvo/selenium-standalone/master

To reproduce issues:
1. Build Docker image:
```
# If exist, delete vvo/selenium-standalone Docker image
git clone https://github.com/vvo/selenium-standalone.git
cd selenium-standalone/
docker build -f Dockerfile -t vvo/selenium-standalone . --rm
docker run -it -p 4444:4444 vvo/selenium-standalone
```

2. From the Selenium web console (http://localhost:4444/wd/hub/static/resource/hub.html):
* Create a Firefox session => Fails with 2 issues:
  * In the web console you see `Unable to create new session.` ribbon
  * In the terminal window: `Error: no DISPLAY environment variable specified`
* Create a Chrome session => Fails with 2 issues:
  * Web console hangs showing loading session tab
  * Terminal window: `[1534427524.313][SEVERE]: bind() returned an error, errno=99: Cannot assign requested address (99)` error message (though it is not the reason of the failure as this message also appears in the fixed Docker image.

Same errors/behaviours occur when running an external program which targets the Selenium server (A simple Python program which works when targeting a selenium-standalone server started locally directly from command line).

**Changes:**
* Group Docker files under `docker` folder (and remove obsolete `.dockerignore` file)
* Do `selenium-standalone install` during image build so application installation is aligned with installed browsers
* Enforce browsers installation from vendor sources (as done in official Selenium Docker images)
* Modify default Chrome startup strategy to enforce usage of parameter `--no-sandbox` (so container does not require `--privileged` parameter (as done in official Selenium Docker images)
* Add `SCREEN_GEOMETRY` environment variable to customize browser window size
* Refactor and comment Dockerfile
* Update documentation
* Include README changes from PR (https://github.com/vvo/selenium-standalone/pull/384)


Dockerfile
```diff
@@ -1,57 +0,0 @@
...
-# For development
-# ADD . ./selenium-standalone-local
-# RUN chown node:node -R .
-USER node
-RUN npm init -y
-# RUN npm install -i ./selenium-standalone-local
-RUN npm install -i selenium-standalone
-
-
-CMD DEBUG=selenium-standalone:* ./node_modules/.bin/selenium-standalone install && DEBUG=selenium-standalone:* ./node_modules/.bin/selenium-standalone start
```
serbanghita: Can you specify on the task what was wrong with this approach or what errors it issued so I can replicate?

docker/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+RUN useradd seluser \
+         --shell /bin/bash  \
+         --create-home \
+  && usermod -a -G sudo seluser \
+  && echo 'ALL ALL = (ALL) NOPASSWD: ALL' >> /etc/sudoers \
+  && echo 'seluser:secret' | chpasswd
+
+
+# Install selenium-standalone
+RUN npm install -g selenium-standalone
```
serbanghita: Previously this was installed as a local account and as a local package to prevent installing any packages as root. Do you know why this technique is better?

catalan-adobe: I do not consider this a better method. The idea here is to shortend and ease commands in the Dockerfile and also, to some degree, emulate how a user would install the application locally.
Also, being in the context of a Docker container, there are no issues installing the application globally (the environment will not changed as we just use the container and then destroy it).

Dockerfile
catalan-adobe: I do not consider the previous approach wrong. I just prefer global installation as it eases understanding of the Dockerfile process. Also it is closer to what a normal user would do locally on its machine.

serbanghita: @catalan-adobe the only problem I had with the current configuration is that it issued a warning about chrome driver, but the service starts and the tests work https://github.com/serbanghita/webdriverio-docker

docker/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
+  && echo "deb http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list \
+  && apt-get update -qqy \
+  && apt-get -qqy install \
+    ${CHROME_VERSION:-google-chrome-stable} \
+  && rm /etc/apt/sources.list.d/google-chrome.list \
+  && rm -rf /var/lib/apt/lists/* /var/cache/apt/*
+
+## Chrome Launch Script Wrapper (forces no-sandbox so no need to use --privileged parameter)
+COPY ./scripts/wrap_chrome_binary /opt/bin/wrap_chrome_binary
```
serbanghita: `Note: --no-sandbox is not needed if you properly setup a user in the container.` [0]
Previously I was adding user `node`

[0] https://developers.google.com/web/updates/2017/04/headless-chrome

Dockerfile
catalan-adobe: What warning?

docker/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+  && echo 'ALL ALL = (ALL) NOPASSWD: ALL' >> /etc/sudoers \
+  && echo 'seluser:secret' | chpasswd
+
+
+# Install selenium-standalone
+RUN npm install -g selenium-standalone
+RUN selenium-standalone install
+
+
+ADD ./scripts/ /home/root/scripts
```
serbanghita: Why `/home/root` and not `/home/seluser`?

docker/Dockerfile
catalan-adobe: I am also adding and using a non-root user (otherwise Chrome would not start). Though, no-sandbox is required (or docker privileged parameter).

docker/Dockerfile
catalan-adobe: No specific reason.

Dockerfile
serbanghita: @catalan-adobe I mean previous configuration: `error in fs.access /home/node/node_modules/selenium-standalone/.selenium/chromedriver/2.40-x64-chromedriver { Error: EACCES: permission denied, access '/home/node/node_modules/selenium-standalone/.selenium/chromedriver/2.40-x64-chromedriver'` - but as far as I investigated this was a race condition in JS which can be fixed with refactori

serbanghita: @catalan-adobe what was not working on the previous image? I ran it and it's starting Selenium on 4444

catalan-adobe: Correct. But then you cannot start browsers.

serbanghita: @catalan-adobe thanks for updating the description about the errors. I'll recheck

docker/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+    xfonts-scalable \
+    xfonts-cyrillic
+
+
+# Install Browser from vendor source
+# Strategies copied from official Selenium Docker images (https://github.com/SeleniumHQ/docker-selenium)
+
+## Install latest Firefox
+ARG FIREFOX_VERSION=latest
+RUN FIREFOX_DOWNLOAD_URL=$(if [ $FIREFOX_VERSION = "latest" ] || [ $FIREFOX_VERSION = "nightly-latest" ] || [ $FIREFOX_VERSION = "devedition-latest" ]; then echo "https://download.mozilla.org/?product=firefox-$FIREFOX_VERSION-ssl&os=linux64&lang=en-US"; else echo "https://download-installer.cdn.mozilla.net/pub/firefox/releases/$FIREFOX_VERSION/linux-x86_64/en-US/firefox-$FIREFOX_VERSION.tar.bz2"; fi) \
```
serbanghita: This is better than the previous 👍

docker/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+  && apt-get update -qqy \
```
serbanghita: Is this line needed again?

docker/Dockerfile
serbanghita: @catalan-adobe To give you some context, the reason why tests previously worked with the current `Dockerfile` is that I ran them in headless mode for both browsers. If you clone `serbanghita/webdriverio-docker` and run `npm test`  and see the configuration at https://github.com/serbanghita/webdriverio-docker/blob/master/wdio.conf.js#L58 , you can see that I've delegated `no-sandbox` there. My choice was not to include this in the container, but to find a workaround (user profile, security profile, etc) and don't run the container privileged or mess with the sandbox.

Overall the wrapper solution you provided is good for now but I would like to provide a better solution in the future.

docker/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+         --shell /bin/bash  \
+         --create-home \
+  && usermod -a -G sudo seluser \
+  && echo 'ALL ALL = (ALL) NOPASSWD: ALL' >> /etc/sudoers \
+  && echo 'seluser:secret' | chpasswd
+
+
+# Install selenium-standalone
+RUN npm install -g selenium-standalone
+RUN selenium-standalone install
```
serbanghita: This better 👍 
I would have preferred to keep the debuging variable for now in case problems occur and we need to debug when we add the CI tests

docker/scripts/wrap_chrome_binary
```diff
@@ -0,0 +1,13 @@
...
+
+WRAPPER_PATH=$(readlink -f /usr/bin/google-chrome)
+BASE_PATH="$WRAPPER_PATH-base"
+mv "$WRAPPER_PATH" "$BASE_PATH"
+
+cat > "$WRAPPER_PATH" <<_EOF
+#!/bin/bash
+
+# Note: exec -a below is a bashism.
+exec -a "\$0" "$BASE_PATH" --no-sandbox "\$@"
```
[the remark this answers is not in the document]
catalan-adobe: Again, this parameter is required to enable spawning Chrome without having to force `--privileged` Docker parameter. 
If we remove it from here, users would have to know about this specific and they will be forced to add it in their configurations.

serbanghita:
`[1534427524.313][SEVERE]: bind() returned an error, errno=99: Cannot assign requested address (99)` - this error doesn't affect the tests, I'm thinking if we ever care about networking we can fix this separately.

catalan-adobe:
@serbanghita, what is your feedback of the error report, could you reproduce the issues?

docker/Dockerfile
catalan-adobe: The provided solution makes both Chrome and Firefox work (headless or not) without users to have to know or configure anything specific.
Also it's the approach currently used by Selenium official Docker images.

docker/scripts/start.sh
```diff
@@ -0,0 +1,7 @@
+
+echo "SCREEN_GEOMETRY: ${SCREEN_GEOMETRY}"
+
+sudo rm -f /tmp/.X*lock
+
+xvfb-run -a --server-args="-screen 0 ${SCREEN_GEOMETRY} -ac +extension RANDR" \
```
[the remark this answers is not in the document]
catalan-adobe: Indeed :)

catalan-adobe:
test

PR #388, opened 2018-08-16
Add Travis stage to build Docker image

catalan-adobe:


catalan-adobe:
Closing PR:
contributed PR too early, still need to be developed.

.travis.yml
```diff
@@ -5,18 +5,28 @@ language: java
...
     - "master"
 
-before_install:
-  - nvm install $TRAVIS_NODE_VERSION
 
-install: "npm install"
+jobs:
+  include:
+
+    - stage: "Build and test application"
```
serbanghita: @catalan-adobe can you include this in https://github.com/vvo/selenium-standalone/pull/387 ? It would be extremely beneficial for future PRs addressing Docker

catalan-adobe: I continued investigating the way to handle update of the Docker image and I don't think Travis is the right place as currently the project is released via `scripts/release.sh`.
I'm currently working on adding the Docker image update+deploy steps in the release script.

PR #389, opened 2018-08-20
Add update of Docker image in release script

catalan-adobe:
* Build latest version of Docker image, tag it and push it to Docker hub for both tags "new release version" and "latest"
* Added comment + minore refactoring
* Get application new version from `package.json` using generic `node` command rather than `json` command


serbanghita:
This looks good. I have a bunch of changes for the release script but I will put them in a separate PR

PR #440, opened 2019-06-21
Fix '--singleDriver' overrides '--driver.<browser>.version'

rhpijnacker:
Using the `--singleDriver` option copies the driver from the `defaultConfiguration`. 
However, if this was overridden by using `--driver...` then that will be lost.
I've updated the code to copy it from the current settings, that include the `--driver...` values.


vvo:
merged in #365

PR #463, opened 2020-07-14
chore: Update binaries versions

### LINKED ISSUE #434: ChromeDriver version
Current driver for chrome `2.43` is kinda bugged and you're unable to maximize the window.
When I changed the version from `2.43` to `2.44` it worked as expected.

Please update the configurations for drivers

just-boris:
fixes #434 

Updated driver versions to the latest available

### DIFF
--- a/test/default-downloads-test.js
+++ b/test/default-downloads-test.js
@@ -184,27 +184,6 @@ describe('default-downloads', function() {
   });
 
   describe('firefox', function() {
-    describe('arm', function() {
-      before(function(){
-        this.originalArch = Object.getOwnPropertyDescriptor(process, 'arch');
-
-        Object.defineProperty(process, 'arch', {
-          value: 'arm'
-        });
-      });
-
-      after(function() {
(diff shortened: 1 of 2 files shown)


test/default-downloads-test.js
```diff
@@ -184,27 +184,6 @@ describe('default-downloads', function() {
   });
 
   describe('firefox', function() {
-    describe('arm', function() {
```
just-boris, line 187: Starting from `0.24.0` release, arm builds are not supported.

vvo: If arm builds were supported previously, how big do you think it's an issue that they won't be supported now? Do you have any link to the source information on that?

Thanks for the PR!

just-boris: There are more details about arm build removal: https://github.com/mozilla/geckodriver/issues/1515

> how big do you think it's an issue 

Also, there were never arm builds for Chromedriver, so I am not sure if there were active users of this setup.

If you think, backward-compatibility is important here, we could have a special condition for arm to download the last available version.

PR #474, opened 2020-09-10
add support for chromium edge

sanyee04:
Add a new browser to the list of supported ones: Chromium Edge.
This change does not affect legacy edge, as there could be some users who might still want to use it.

The webdriver unzipping process was also slightly updated to support zips which might contain folders and not just the webdriver binary (which is the case for the latest chromium edge drivers)

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -174,6 +174,14 @@ function install(opts, cb) {
       })
     }
 
+    if (opts.fsPaths.chromiumedge) {
+      installers.push({
+        installer: installChromiumEdgeDr,
+        from: opts.urls.chromiumedge,
+        to: opts.fsPaths.chromiumedge.downloadPath
+      })
+    }
+
     var steps = installers.map(function (opts) {
       return onlyInstallMissingFiles.bind(null, opts);
     });
@@ -290,6 +298,10 @@ function install(opts, cb) {
     }
   }
 
+  function installChromiumEdgeDr(opts, cb) {
+    installZippedFile(opts.from, opts.to, cb);
+  }
+
   function installGzippedFile(from, to, cb) {
     getDownloadStream(from, function(err, stream) {
       if (err) {
@@ -397,13 +409,15 @@ function install(opts, cb) {
     var yauzl = require('yauzl');
     var extractPath = path.join(path.dirname(zipFilePath), path.basename(zipFilePath, '.zip'));
 
-    yauzl.open(zipFilePath, {lazyEntries: true}, function onOpenZipFile(err, zipFile) {
+    yauzl.open(zipFilePath, function onOpenZipFile(err, zipFile) {
       if (err) {
         cb(err);
         return;
       }
-      zipFile.readEntry();
-      zipFile.once('entry', function (entry) {
+      zipFile.on('entry', function (entry) {
+        if (/.*\/.*/.test(entry.fileName)) {
+          return; // ignore folders, i.e. release notes folder in edge driver zip
+        }
         zipFile.openReadStream(entry, function onOpenZipFileEntryReadStream(err, readStream) {
           if (err) {
             cb(err);
@@ -487,7 +501,7 @@ function setDriverFilePermissions(where, cb) {
 }
 
 function logInstallSummary(logger, paths, urls) {
-  ['selenium', 'chrome', 'ie', 'firefox', 'edge'].forEach(function log(name) {
+  ['selenium', 'chrome', 'ie', 'firefox', 'edge', 'chromiumedge'].forEach(function log(name) {
     if (!paths[name]) {
       return;
     }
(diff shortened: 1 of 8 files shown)


lib/install.js
```diff
@@ -397,13 +409,15 @@ function install(opts, cb) {
     var yauzl = require('yauzl');
     var extractPath = path.join(path.dirname(zipFilePath), path.basename(zipFilePath, '.zip'));
 
-    yauzl.open(zipFilePath, {lazyEntries: true}, function onOpenZipFile(err, zipFile) {
```
vvo, line 400: Can you comment on why this was removed and what's the impact on other drivers?

lib/install.js
```diff
@@ -397,13 +409,15 @@ function install(opts, cb) {
...
+    yauzl.open(zipFilePath, function onOpenZipFile(err, zipFile) {
       if (err) {
         cb(err);
         return;
       }
-      zipFile.readEntry();
-      zipFile.once('entry', function (entry) {
+      zipFile.on('entry', function (entry) {
+        if (/.*\/.*/.test(entry.fileName)) {
```
vvo, line 418: Could this have a negative impact on other drivers?

test/no-reinstall.js
```diff
@@ -1,6 +1,8 @@
 describe('when files are installed', function () {
     it('should not reinstall them', function (done) {
 
+        this.timeout(120000);
```
vvo, line 4: Why did you have to add this line while previously it was not present? Thanks

vvo (commented):
Added some questions

vvo:
Can you also add some documentation (README) so people know that edge driver is now supported (+ add note about the differences between old edge and new edge).

vvo:
- [ ] any issues/PRs we can close following support of new edge driver?
- [ ] should we remove legacy edge support at some point?

lib/install.js
sanyee04: @vvo the previous code took the first file it found inside the zip and just unzipped it in the destination folder. This worked fine for all other webdrivers as the downloaded zip for those contained nothing else, but the webdriver binary.
Unfortunately, for the chromium edge driver this is not the case, as the zip has this structure:
```
Driver_Notes
-- | credits.html
-- | LICENSE
msedgedriver.exe
```
The previous code, because it only unzipped the first file it found, kept unzipping the `Driver_Notes/credits.html` file and it `selenium-standalone start` obviously failed as an html file can't be used as a webdriver.

The changes have no effect for other webdrivers, what the new code does is ignore all folders inside the zip, and unzip the first file inside it (so the other webdrivers' zip structure works as well)

lib/install.js
sanyee04: Not that I'm aware of, I looked into the `yauzl` documentation, their guides state that this is a valid check for folders inside zip files. The only negative side-effect would be if some browsers suddenly start putting their webdrivers inside folders in the zip, but if that happens we'd need to make some changes in the code anyway.

test/no-reinstall.js
sanyee04: Unfortunately, the CDN which serves the chromium edge driver takes a bit longer to download the zip sometimes, so I had to increase the test timeout in order to make it always pass. Usually it should take a few seconds only but I wanted to make this change so future changes will also make all tests pass.

lib/install.js
vvo: 👍

lib/install.js
vvo: 👍

test/no-reinstall.js
vvo: 👍

vvo:
@sanyee04 I added a few check items for you, then we should be good to move and release :)

sanyee04:
> @sanyee04 I added a few check items for you, then we should be good to move and release :)

Thanks! I'll make the required changes before the end of the day and update the PR!

sanyee04:
The following issues can be closed after this fix is merged:
- https://github.com/vvo/selenium-standalone/issues/458
- https://github.com/vvo/selenium-standalone/issues/459

sanyee04:
@vvo legacy edge end of life is **March 9, 2021**, based on [this](https://techcommunity.microsoft.com/t5/microsoft-365-blog/microsoft-365-apps-say-farewell-to-internet-explorer-11-and/ba-p/1591666?WT.mc_id=modinfra-0000-abartolo).
I say we should keep legacy edge support until then, and we can remove it afterwards.

vvo:
Thanks @Petersandor-calvi too :)

sanyee04:
> Thanks @Petersandor-calvi too :)

Yeah that's my job account, got confused and pushed the commit with that one :)
Thanks for merging this @vvo, we really needed this!

PR #476, opened 2020-09-11
Added EdgeDriverSupportForChromium

Manikanta-Munnangi:
Hi, Checkout the implementation for chromium edge driver. which, i removed the legacy version of edge completely.
and it will detect the installed edge browser version and install the matching driver only when browser have installed in default location.  Also, you can install by overdriving the version variable.

Please, ignore the extra spaces which are trimmed by format document. git sensed the spaces also add as changes. Thanks !


vvo:
hey @Manikanta-Munnangi there's an ongoing PR that was opened slightly before yours, see: #474 wdyt?

Manikanta-Munnangi:
Yeah! i seen that now :(

vvo:
Closing this, sorry that both of you implemented it, thanks for the big effort though!

Manikanta-Munnangi:
@vvo  #474 this will get error when we trying to override the edge chromium version from webdriverIO, services i mean from install args. Even though it will install correct edge chromium driver. it will fetch version which is harded-coded in default.config file.

Manikanta-Munnangi:
@vvo #474 this is only working above chromium edge 86 version and when i try to override the chromiumedge version with my browser-driver version from webdriverio services i mean from install args like below.
OverrideEdgeChromiumVersion

it says this 

Screenshot (112)

my approach was different will work even when you override the version.

PR #484, opened 2020-10-23
ESlint with airbnb rules

marlonemmerich:


mgrybyk:
Thanks for the PR! 
It's important to run linter in CI and make it pass.
Maybe it would be easier to start with less strict rules?
Additionally, it would be great to run lint on pre commit hook with husky.

package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "start": "DEBUG=selenium-standalone:* ./bin/selenium-standalone install && DEBUG=selenium-standalone:* ./bin/selenium-standalone start",
     "test": "./bin/selenium-standalone install && mocha",
+    "lint": "eslint bin/** lib/**",
```
mgrybyk, line 9: Let's check js files only

marlonemmerich:
Okay, sure! I will work on this

marlonemmerich:
Hi! I'm sorry, I'm late =\
About run linter in CI, i'm sorry., you told my and i forgot.
I'm not sure if I can test it here (i'm front end haha), can you check it's OK now?

In lint rules, i put this:

`
"no-irregular-whitespace": "error",
`
`
"no-alert": "error",
`
`
"no-loop-func": "error",
`
`
"array-callback-return": "error"
`

It's OK in this moment? Do you have any suggestions?

Thanks!

mgrybyk:
I'll raise a PR to deal with that, thanks a lot for your efforts!

PR #518, opened 2020-11-30
Fix 449

jonesh66:
When Selenium 4 is being used, 
1. Change path to `/status`.
2. Append `standalone` to seleniumArgs


lib/default-config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = () => {
   const config = {
     baseURL: 'https://selenium-release.storage.googleapis.com',
-    version: '3.141.59',
+    version: '4.0.0-alpha-3',
```
mgrybyk, line 4: Please revert this change

jonesh66:
@mgrybyk will add tests for this change as soon as I figure out where to add them

lib/get-selenium-status-url.js
```diff
@@ -68,7 +69,7 @@ exports.getSeleniumStatusUrl = function (seleniumArgs) {
   }
 
   const statusURI = new URL('http://' + host);
-  const nodeStatusAPIPath = '/wd/hub/status';
+  const nodeStatusAPIPath = defaultConfig.version.startsWith('4.') ? '/status' : '/wd/hub/status';
```
mgrybyk, line 72: Don't use default config to determine version

lib/start.js
```diff
@@ -32,6 +32,10 @@ function start(_opts, _cb) {
     opts.spawnCb = noop;
   }
 
+  if (defaultConfig.version.startsWith('4.')) {
```
mgrybyk, line 35: Don't use default config for version detection

mgrybyk (changes requested):

lib/start.js
```diff
@@ -32,6 +32,10 @@ function start(_opts, _cb) {
...
+    opts.seleniumArgs.push('standalone');
```
mgrybyk, line 36: Only push it if it's not already there

lib/start.js
jonesh66: using opts now

lib/get-selenium-status-url.js
jonesh66: using opts here

lib/default-config.js
mgrybyk: This one still should be reverted

lib/get-selenium-status-url.js
```diff
@@ -35,6 +35,8 @@ exports.getSeleniumStatusUrl = function (seleniumArgs) {
   const nodeConfigArg = seleniumArgs.indexOf('-nodeConfig');
   const portArg = seleniumArgs.indexOf('-port');
   const hostArg = seleniumArgs.indexOf('-host');
+  const isSelenium4 = seleniumArgs[4];
```
mgrybyk, line 38: Please don't rely on index

mgrybyk (changes requested):

lib/get-selenium-status-url.js
jonesh66: what argument could be used to determine the version? 
I want to use something like `const isSelenium4 = seleniumArgs.indexOf('-version')`

lib/start.js
```diff
@@ -32,6 +32,10 @@ function start(_opts, _cb) {
     opts.spawnCb = noop;
   }
 
+  if (opts.version.startsWith('4.') && opts.seleniumArgs.indexOf('standalone') === -1) {
```
mgrybyk, line 35: I forgot to take into account that we should only add standalone if the mode is not specified explicitly.
https://www.automatetheplanet.com/selenium-webdriver-4-overview/
```
// add somewhere to the top
const seleniumModes = ['standalone', 'hub', 'node']

// replace opts.seleniumArgs.indexOf('standalone') === -1 with
!seleniumModes.some((a) => seleniumArgs.includes(a))
```

lib/get-selenium-status-url.js
mgrybyk: let's pass `opts` to `checkStarted` function, then pass it to `getSeleniumStatusUrl`.

Don't use args to detect version, use opts instead

mgrybyk (changes requested):

lib/start.js
mgrybyk: Hm, I think it would be even better if at least one arg passed and that's it. 
```suggestion
  if (opts.version.startsWith('4.') && opts.seleniumArgs.length === 0) {
```

lib/get-selenium-status-url.js
jonesh66: I haven't been able to figure out a way to pass opts to checkstarted.
`async function checkStarted(seleniumArgs, opts) {...`
how do I know what value does opts hold here? I think it has something to do with `checkArgs` but I am not sure how to proceed. Could you please advise?

mgrybyk: you can use debugger and follow the call chain to figure out which function is calling checkStarted function

image

`checkPathsExistence` -> `checkStarted`

so, you have to pass opts to checkPathsExistence, then pass opts from checkPathsExistence to checkStarted

mgrybyk: following that I'm able to pass opts
image

mgrybyk: here you go https://github.com/vvo/selenium-standalone/compare/pass-opts-to-check-started

mgrybyk:
@jonesh66 do you need any help?

jonesh66:
@mgrybyk I think I'm using opts now but the tests are failing

mgrybyk:
@jonesh66 see logs, looks like you are trying to get version from undefined.
Please fix tests

mgrybyk:
Can I help with something else here?

lib/default-config.js
mgrybyk: Ping

jonesh66: sorry I was still testing this using the config. I'll revert this to the original once the tests pass. (i know it's messy)

PR #529, opened 2020-12-11
update chromiumedge default URL

krulletc:
The base URL for Microsoft's Edge Chrome driver appears to have been changed by Microsoft. This fix simply updates that URL.

To verify, the new list of edge driver files is viewable at: https://msedgedriver.azureedge.net/

### DIFF
--- a/lib/default-config.js
+++ b/lib/default-config.js
@@ -27,7 +27,7 @@ module.exports = () => {
         version: 'latest',
         fallbackVersion: '87.0.637.0',
         arch: process.arch,
-        baseURL: 'https://msedgedriver.azureedge.net',
+        baseURL: 'https://msedgewebdriverstorage.blob.core.windows.net/edgewebdriver',
       },
     },
   };


mgrybyk:
I don't think we need to change the url unless the current one works fine

krulletc:
@mgrybyk - From what I can tell Microsoft did not upload the mac version to release 89, so now downloads are broken on Mac.  A mac user will receive a 404.  This update changes the URL to their Azure Blob that does contain the mac versions.  If there's a solution that you would prefer over using this URL, then I can make it.  However, for now Mac users will receive a 404 when getting the Edge Driver download with a clean install of the latest `selenium-standalone`

mgrybyk:
I'm not getting any errors

mgrybyk:
I tried:
- default args `https://msedgedriver.azureedge.net/87.0.664.60/edgedriver_mac64.zip` - OK
- `selenium-standalone install --drivers.chromiumedge.version=89.0.727.0` - `https://msedgedriver.azureedge.net/89.0.727.0/edgedriver_mac64.zip` - OK

krulletc:
Interesting, it is working for me this morning, as well.  Last night, the https://msedgedriver.azureedge.net/87.0.664.60/edgedriver_mac64.zip url always returned a 400.  Perhaps there was an Azure issue.  I will close this issue.  Sorry for the confusion.

mgrybyk:
No problem. There are quite often some issues with edge driver download, sorry to say you've faced one of them.

Anyway, thanks for the investigation. In case the current URL stops working we can just reopen the PR

PR #544, opened 2021-03-03
Fix "chromium" typo

arniscg:
There is a typo in following error message
```
Only x64 architecture is available for chroimum edge driver for linux
```


mgrybyk (approved):
👍 
Thank you!

PR #562, opened 2021-06-30
Fix open handle in install-utils

relu91:
I was using selenium with jest and I got complaints about an open handle in install-utils. My hypothesis is that the test setup that I am using does not use `uncompressGzippedFile` function and the gunzip stream is never used/closed. Moving the creation of the gunzip stream inside the method fixed my problem.

### DIFF
--- a/lib/install-utils.js
+++ b/lib/install-utils.js
@@ -5,6 +5,7 @@ const mkdirp = require('mkdirp');
 const path = require('path');
 const yauzl = require('yauzl');
 const fs = require('fs');
+const zlib = require('zlib');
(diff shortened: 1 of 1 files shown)


christian-bromann:
@relu91 thanks for the contribution. Can we keep the import still at the top of the file?

relu91:
> @relu91 thanks for the contribution. Can we keep the import still at the top of the file?

Yes, you are right it makes more sense with the current codestyle. Just pushed the change now 👍🏻

christian-bromann (approved):
Thanks, lgtm 👍

PR #591, opened 2021-11-25
Add message for edgedriver download failures

harsha509:
As edge driver is unavailable for a few versions on fewer platforms, Webdriverio fails to start the server with 404 messages.

This PR is to print a precise message in case if this happens.
**Note**: This issue only occurs for chromiumedge webdriver downloads.

HI @christian-bromann,

May not be perfect, but let me know if this works!

Screenshot 2021-11-25 at 8 29 34 PM

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -301,7 +301,21 @@ async function install(_opts) {
         .once('end', () => {
           downloadStreams.delete(downloadStream);
         })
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):
:thumbsup: thanks a lot!

Issue #620, opened 2022-02-15
Linux chrome driver downloads msedgedriver not avaliable for linux


senpl:
## Expected Behavior
I just want to install chrome driver for linux.
## Current Behavior
It fails with msg:
`ENOENT: no such file or directory, chmod '/home/l/.npm-packages/lib/node_modules/selenium-standalone/.selenium/chromiumedgedriver/latest-x64/msedgedriver'`
I do not even want msedgedriver.
## Steps to Reproduce
selenium-standalone install --drivers.chrome.version=98.0.4758.80 --drivers.chrome.baseURL=https://chromedriver.storage.googleapis.com

1.  download google-chrome
2.  install selenium-standalone
3.  selenium-standalone install --drivers.chrome.version=98.0.4758.80 --drivers.chrome.baseURL=https://chromedriver.storage.googleapis.com

## Logs
```
Error in "chmod". 
See more details below:
ENOENT: no such file or directory, chmod '/home/l/.npm-packages/lib/node_modules/selenium-standalone/.selenium/chromiumedgedriver/latest-x64/msedgedriver'
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^

[Error: ENOENT: no such file or directory, chmod '/home/l/.npm-packages/lib/node_modules/selenium-standalone/.selenium/chromiumedgedriver/latest-x64/msedgedriver'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'chmod',
  path: '/home/l/.npm-packages/lib/node_modules/selenium-standalone/.selenium/chromiumedgedriver/latest-x64/msedgedriver'
}

Node.js v17.3.0
```

If there are lots of logs, please link to [a public gist](https://gist.github.com/).

## Your Environment
* Version of `selenium-standalone` that you are using: 8.0.9
* Is there another tool calling `selenium-standalone` on your behalf:no
* System/platform:Arch Linux 
* Other details that might be important:might be related to https://github.com/webdriverio/webdriverio/issues/8006 still it's not same, coz here driver probably not exist, and no known fix exist afaik

christian-bromann:
Thanks for reporting!

Any contributions that resolves the bug are highly appreciated. Please take a look into our [contribution guidelines](https://github.com/webdriverio/selenium-standalone/blob/main/CONTRIBUTING.md) and let us know if you have any questions. Cheers!

cshuiu:
[...]

PR #623, opened 2022-02-23
fix for promises to fix issue with msedge driver

senpl:
should fix https://github.com/webdriverio/selenium-standalone/issues/620#issuecomment-1048808131
or at least fix build on linux


lib/install-utils.js
```diff
@@ -5,6 +5,7 @@ const mkdirp = require('mkdirp');
 const path = require('path');
 const yauzl = require('yauzl');
 const fs = require('fs');
+const fsPromises = require('fs').promises;
```
christian-bromann, line 8:
```suggestion
const fsPromises = require('fs/promises');
```

lib/install-utils.js
```diff
@@ -66,20 +67,21 @@ const chmod = (where) =>
...
-    })
-  );
+  const requireChmod = await gotAccess(where)
 
   if (requireChmod) {
     await chmod(where);
   }
 }
 
+async function gotAccess(where) {
```
christian-bromann, line 77: This function will always return `undefined`, let's make sure we return something 😉

christian-bromann (changes requested):

christian-bromann:
@senpl can you check there seem to be a conflict in `lib/install-utils.js`

lib/install-utils.js
senpl: I added returned value. it works on my machine.

christian-bromann:
@senpl why do you think this fixes the issue we have in Linux? The current PR rewrites the same behavior just moves it out into a function.

senpl:
> @senpl why do you think this fixes the issue we have in Linux? The current PR rewrites the same behavior just moves it out into a function.

When I submitted pull request it was not done in async way, and error from permission was send before other chcecks. When awaits are used error with file not downloaded seems to not be present on my machine. Still issue not got test case so it's hard to tell if this is what fix an issue. Anyway I wanted to fix issue in times when access check was not done in async way.

senpl:
> @senpl why do you think this fixes the issue we have in Linux? The current PR rewrites the same behavior just moves it out into a function.

Also I am the one who report it. So If I get underline issue adding awaits fix part that not waiting for extraction a file. I checked latest version and the issue is fixed, so probably this merge is not needed. Still it fix issue for older version for me. So I could close it, as it's not really crucial now.

christian-bromann:
Closing then. Thanks for looking into this anyway.

PR #629, opened 2022-03-08
Remove callback and replace with promise

harsha509:
Removes callback in install-utils `checksum`


christian-bromann (approved):
Promises FTW 🎉

Issue #653, opened 2022-04-19
[🐛 Bug]: error in downloading selenium-standalone - Error in "getDownloadStream" #592


wyhmichael:
### Have you read the Contributing Guidelines on issues?
 * [x]  I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).
 
 ### WebdriverIO Version
 latest
 
 ### Node.js Version
 latest
 
 ### Mode
 Standalone Mode
 
 ### Which capabilities are you using?
 ```ts
 {
             browserName: 'chrome',
             acceptInsecureCerts: true,
             'goog:chromeOptions': {
                 args: ['--headless', '--disable-gpu', "--no-sandbox", '--window-size=1280,800', '--disable-dev-shm-usage', '--disable-web-security']
 }
 
         installArgs: {
             version : "3.141.59",
             baseURL : "https://selenium-release.storage.googleapis.com",
             drivers : {
                 chrome : {
                     version : "95.0.4638.69",
                     arch    : process.arch,
                     baseURL : "https://chromedriver.storage.googleapis.com"
                 }
             }
         }
 ```
 
 ### What happened?
 from last few days, i started getting below error `Error in "getDownloadStream". Could not download https://selenium-release.storage.googleapis.com/selenium-3.141.59/selenium-server-standalone-3.141.59.jar 442See more details below: 443404 https://selenium-release.storage.googleapis.com/selenium-3.141.59/selenium-server-standalone-3.141.59.jar 444Response code 404 (Not Found)`
 
 the url seems invalid. correct url is https://selenium-release.storage.googleapis.com/3.141/selenium-server-standalone-3.141.59.jar. how to change it? ` installArgs: { version : "3.141.59", baseURL : "https://selenium-release.storage.googleapis.com", drivers : { chrome : { version : "95.0.4638.69", arch : process.arch, baseURL : "https://chromedriver.storage.googleapis.com" } } },`
 
 ### What is your expected behavior?
 selenium should be downloaded without any issues.
 
 ### How to reproduce the bug.
 start tests with the below config
 ```
 config.services = [
      [
        'selenium-standalone',
        {
          logPath: 'logs/selenium',
          installArgs: {
              version: '3.141.59',
              baseURL: 'https://selenium-release.storage.googleapis.com'
              chrome: {
                version: '99.0.4844.51',
                arch: process.arch,
[...]

PR #654, opened 2022-04-19
fix https://github.com/webdriverio/selenium-standalone/issues/653

wyhmichael:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "selenium-standalone",
-  "version": "8.0.10",
+  "version": "8.0.11",
```
christian-bromann, line 3: We have an automated way to update the version
```suggestion
  "version": "8.0.10",
```

christian-bromann (changes requested):
Can we add a test for Selenium 4?

wyhmichael:
> Can we add a test for Selenium 4?

Thanks @christian-bromann  I reverted the version update

And there's tests for Selenium 4, and looks good already: https://github.com/webdriverio/selenium-standalone/pull/654/files#diff-96299f8ea73dbd04f0260ab3ae2db9a749fbd43498bd9733807d2c40405d9bf6L65-R108, do you want more tests around it?

christian-bromann (approved):
Ah ok, in this case this looks good to me 👍 

Thanks for the contribution!

wyhmichael:
> Ah ok, in this case this looks good to me 👍
> 
> Thanks for the contribution!

Thanks @christian-bromann for confirmation!

Sorry that I had a linting error, would you pls take another look?

PR #655, opened 2022-04-23
docs: update basePath documentation

### LINKED ISSUE #631: basePath documentation inconsistent with behavior
## Expected Behavior
This likely wouldn't impact most people, but the documentation for opts.basePath is inconsistent with the actual behavior.

[Documentation](https://github.com/webdriverio/selenium-standalone/blob/main/docs/API.md#seleniuminstallopts) states basePath default value is `process.cwd() + ".selenium"` while the programmed behavior is relative to [`__dirname`](https://github.com/webdriverio/selenium-standalone/blob/102dd8d78277942dc4f560129d2b340627498ef3/lib/compute-fs-paths.js#L5).

Like I said, most people wouldn't notice but I happen to have a project where `__dirname` and `process.cwd()` values are very different.

If desired, i can make the change if someone would let me know which to follow, meaning either the code is correct and the documentation needs updating or the documentation is correct and the code needs a fix.

is2ei:
fixes #631

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -78,7 +78,7 @@ arch [sometimes](https://code.google.com/p/selenium/issues/detail?id=5116#c9).
 
 `opts.ignoreExtraDrivers` only downloads and installs drivers explicity specified. Broken https://github.com/webdriverio/selenium-standalone/issues/421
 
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
👍 thanks!

Issue #617, opened 2022-02-02
Issue in check-started.js function as it does not wait for 1 minute to start the server


mighty98:
## Expected Behavior
As per the comments in check-started.js function (Line 37) server should be allowed 1 minute to start but as per the logic it wait for just 15 seconds

`let attempts = 29;
  const startTime = Date.now();
  while (attempts > 0 && Date.now() - startTime < 30000) {
    await sleep(500);
    attempts--;
    if (cpState.exited) {
      throw new Error(
        `Selenium exited before it could start with code ${cpState.code}\nStdout: ${cpState.stdout}\nStderr: ${cpState.stderr}`
      );
    }
    try {
      // server has one minute to start
      await got(seleniumStatusUrl, options);
      selenium.removeListener('exit', errorIfNeverStarted);
      return null;
    } catch (err) {
      // suppress some messages
      if (attempts % 5 === 0) {
        console.error('Failed to connect to selenium.', 'Attempts left:', attempts, '\n', err.message);
      }
    }
}`

## Current Behavior
As per above logic we wait actually just 15 seconds for the server to start
Instead shouldn't the code be as below (wait max 60 seconds and check every 2 seconds)
` let attempts = 29;
  const startTime = Date.now();
  while (attempts > 0 && Date.now() - startTime < 60000) {
    await sleep(2000);
    attempts--;
`

## Your Environment
* Version of `selenium-standalone` that you are using: 8.0.9 
* Is there another tool calling `selenium-standalone` on your behalf: @wdio/selenium-standalone-server
* System/platform:Windows/Ubuntu
* Other details that might be important:

christian-bromann:
Thanks for reporting!



Any contributions that resolves the bug are highly appreciated. Please take a look into our [contribution guidelines](https://webdriver.io/docs/contribute) and let us know if you have any questions. Cheers!

udarrr:
It has been resolved in https://github.com/webdriverio/selenium-standalone/pull/668

PR #662, opened 2022-04-29
fixed timeout for server started checks #617

oleksiizozulenko:
Also added possibility customize timeouts via env variables  SELENUIM_CHECK_STARTED_TIMEOUT and DEFAULT_SELENUIM_CHECK_STARTED_PAUSE


christian-bromann:
Can you try to squash them all into one commit with the right email? Otherwise another option would be to just create a new PR.

PR #664, opened 2022-04-29
Fix Selenium URLs prior to v4

### LINKED ISSUE #659: Won't Download Older Versions of Selenium
## Expected Behavior

When specifying a version below v4 of Selenium the install task should download that older version of Selenium.

## Current Behavior

Currently it generates the wrong URL and therefore resulting in a 404.

The issue is two things:

* [The stripping of the patch version number](https://github.com/webdriverio/selenium-standalone/blob/main/lib/compute-download-urls.js#L49) when determining part of the path. It appears the default base URL [used to be](https://github.com/webdriverio/selenium-standalone/commit/500585bb008a0caf09a904f37a5e1f91cdbb2100) https://selenium-release.storage.googleapis.com while now it is https://github.com/SeleniumHQ/selenium/releases/download. The old base URL did strip the patch version number while the new base URL does not.
* The `selenium-` prefix in that same path component is also now part of the location.

We can restore access to most of the selenium versions by replacing [this line](https://github.com/webdriverio/selenium-standalone/blob/main/lib/compute-download-urls.js#L49) with:

```js
          : `selenium-${opts.seleniumVersion}`
```

With this change there are still a handful of versions that cannot be downloaded (such as 3.150.0) but that is because those versions do not include a pre-built jar and therefore there is nothing we can do about that. If this change sounds good I can put together a quick PR.

## Steps to Reproduce

Given this `package.json`:

```json
{
  "name": "ssa-test",
  "private": true,
  "type": "module",
  "devDependencies": {
    "selenium-standalone": "^8.0.11"
  }
}
```

And this usage of the install API:

```js
#!/usr/bin/env node

import selenium from 'selenium-standalone'

;(async function() {
  await selenium.install({ version: '3.141.0' })
})()
```

I get this error:

```
Error in "getDownloadStream". Could not download https://github.com/SeleniumHQ/selenium/releases/download/3.141/selenium-server-standalone-3.141.0.jar
See more details below:
404 https://github.com/SeleniumHQ/selenium/releases/download/3.141/selenium-server-standalone-3.141.0.jar
Response code 404 (Not Found)
/home/me/tmp/ssa-test/node_modules/selenium-standalone/lib/install.js:318
          throw new Error('Could not download ' + downloadUrl);
          ^

Error: Could not download https://github.com/SeleniumHQ/selenium/releases/download/3.141/selenium-server-standalone-3.141.0.jar
    at Request.<anonymous> (/home/me/tmp/ssa-test/node_modules/selenium-standalone/lib/install.js:318:17)
    at Object.onceWrapper (node:events:642:26)
    at Request.emit (node:events:527:28)
    at emitErrorNT (node:internal/streams/destroy:151:8)
    at emitErrorCloseNT (node:internal/streams/destroy:116:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)

Node.js v18.0.0
```

## Logs

I wasn't sure which versions of selenium could and could not be downloaded. So I put together a quick script to use this libaries functions to generate the URL and then do a test `HEAD` request. The below is that script:

```js
#!/usr/bin/env node

// Prints all versions of selenium which selenium-standalone cannot compute
// a valid selenium version URL.

import computeDownloadUrls from 'selenium-standalone/lib/compute-download-urls.js'
import defaultConfig from 'selenium-standalone/lib/default-config.js'
import got from 'got'

const defaults = defaultConfig()
const versions = [
  '2.53.1', '2.53.0', '2.52.2', '2.52.1', '2.52.0', '2.51.0', '2.50.1',
  '2.50.0', '2.49.1', '2.49.0', '2.48.2', '2.48.1', '2.48.0', '2.47.1',
  '2.47.0', '2.46.0', '2.45.0', '2.44.0', '2.43.1', '2.43.0', '2.42.2',
  '2.42.1', '2.42.0', '2.41.0', '2.40.0', '2.39.0', '3.0.0-beta-1',
  '3.0.0-beta-2', '3.0.0-beta-3', '3.0.0-beta-4', '3.0.0', '3.0.1', '3.1.0',
  '3.2.0', '3.3.0', '3.3.1', '3.4.0', '3.5.0', '3.5.1', '3.5.2', '3.5.3',
  '3.6.0', '3.7.0', '3.7.1', '3.8.0', '3.8.1', '3.9.0', '3.9.1', '3.10.0',
  '3.11.0', '3.12.0', '3.13.0', '3.14.0', '3.141.0', '3.141.5', '3.141.59',
  '3.150.0', '4.0.0-alpha-1', '4.0.0-alpha-2', '4.0.0-alpha-3', '4.0.0-alpha-4',
  '4.0.0-alpha-5', '4.0.0-alpha-6', '4.0.0-alpha-7', '4.0.0-beta-1',
  '4.0.0-beta-2', '4.0.0-beta-3', '4.0.0-beta-4', '4.0.0-rc-1', '4.0.0-rc-2',
  '4.0.0-rc-3', '4.0.0', '4.1.0',
]

async function checkVersion(version) {
  const urls = await computeDownloadUrls({
    seleniumVersion: version,
    seleniumBaseURL: defaults.baseURL,
    drivers: {}
  });

  const { statusCode } = await got(
    urls.selenium,
    { method: 'HEAD', throwHttpErrors: false }
  )
  return statusCode == 200
}

;(async function() {
  Promise.all(versions.map(async version => {
    if( !await checkVersion(version) ) console.log(version)
  }))
})()
```

Using the existing code only v4 versions of Selenium work. With my suggested change the only output is:

```
2.52.2
2.52.1
3.0.0-beta-1
3.0.0-beta-2
3.0.0-beta-3
3.0.0-beta-4
3.150.0
```

Those versions are failing because there is not pre-built jar file so nothing we can do about that. It might be worth considering incorporating a version of this script into the repo to run occasionally to verify the URLs. We could ignore the versions we know that will fail and also maybe verify the URLs for the web drivers.

## Your Environment
* Version of `selenium-standalone` that you are using: 8.0.11
* Is there another tool calling `selenium-standalone` on your behalf: Hit the issue while working with PouchDB but my reproduction case shows it is independent of that tool.
* System/platform: Linux (PopOS)

eric1234:
This should fix #659.

I divided this up into two commits. The first is just the simple fix along with an update to the unit tests to match.

But obviously the tests were passing just fine previously even though the URLs were invalid. Therefore the second commit ups our testing so we are doing a smoke test against the URL for every released version (according to Github). This should prevent a similar regression in the future.

Since this test uses the internet I did make it bail out and skip the test if it can't get that release list under the idea that the developer is disconnected. I had to increase the timeout a bit for that test. On my machine it took a bit over 2 seconds to get that release list and check each URL. 5 seconds should give padding if someone has a slow connection.

If this sort of integration test is not something you want in the test suite feel free to pop the commit off via a rebase and just merge the first one.

Also I did make sure to run the linter and updated my syntax to make the nanny happy so hopefully should avoid the issues I got last time.

### DIFF
--- a/lib/compute-download-urls.js
+++ b/lib/compute-download-urls.js
@@ -46,7 +46,7 @@ async function computeDownloadUrls(opts) {
         opts.seleniumBaseURL,
         isSelenium4(opts.seleniumVersion)
           ? `selenium-${getVersionWithZeroedPatchPart(opts.seleniumVersion)}`
-          : opts.seleniumVersion.replace(/(\d+\.\d+)\.\d+/, '$1'),
+          : `selenium-${opts.seleniumVersion}`,
         opts.seleniumVersion
       ),
   };
--- a/test/compute-download-urls-test.js
+++ b/test/compute-download-urls-test.js
@@ -1,4 +1,6 @@
 const assert = require('assert');
+const got = require('got');
+const defaults = require('../lib/default-config.js')();
 
 let computeDownloadUrls;
 
@@ -56,7 +58,10 @@ describe('compute-download-urls', () => {
         drivers: {},
       });
 
-      assert.strictEqual(actual.selenium, 'https://localhost/selenium-3.0.0-beta2/selenium-server-standalone-3.0.0-beta2.jar');
+      assert.strictEqual(
+        actual.selenium,
+        'https://localhost/selenium-3.0.0-beta2/selenium-server-standalone-3.0.0-beta2.jar'
+      );
     });
 
     it('version 4 basic', async () => {
@@ -101,6 +106,38 @@ describe('compute-download-urls', () => {
         actual.selenium,
         'https://selenium-release.storage.googleapis.com/4.0-alpha-7/selenium-server-4.0.0-alpha-7.jar'
       );
+    });
+
+    it('generates URLs that respond successfully', async function () {
+      this.timeout(5000); // HTTP requests take a few seconds
+
+      const versionsExpectedToFail = ['3.150.0'];
+
+      let data;
+      try {
+        const releasesURL = 'https://api.github.com/repos/SeleniumHQ/selenium/releases';
+        data = await got(releasesURL).json();
+      } catch (e) {
+        // Likely no internet connection so skip but output error to help
+        // debug in case something else.
+        console.debug(e);
+        return this.skip();
+      }
+
+      const versions = data.map((release) => release.tag_name.replace(/^selenium-/, ''));
+      const checks = versions.map(async (version) => {
+        if (versionsExpectedToFail.includes(version)) return;
+
+        const urls = await computeDownloadUrls({
+          seleniumVersion: version,
+          seleniumBaseURL: defaults.baseURL,
+          drivers: {},
+        });
+
+        const { statusCode } = await got(urls.selenium, { method: 'HEAD', throwHttpErrors: false });
+        assert.strictEqual(statusCode, 200, `URL for Selenium ${version} does not look valid`);
+      });
+      await Promise.all(checks);
     });
   });
 
--- a/test/compute-download-urls-test.js
+++ b/test/compute-download-urls-test.js
@@ -36,7 +36,7 @@ describe('compute-download-urls', () => {
         drivers: {},
       });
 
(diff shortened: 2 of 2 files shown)


christian-bromann (approved):
👍 thanks for the great PR description and the fix!

PR #668, opened 2022-05-03
bug #617 improve timeouts for GOT utils

oleksiizozulenko:
After more details testing found that timeout for got utils and timeout for startup checks should be different. 
Also in gotOptions timeout should be `object` according to [documentation](https://github.com/sindresorhus/got/blob/6f5c7ce1233c7a28fed12d64766fa19e8eb3b345/documentation/6-timeout.md)


christian-bromann (approved):
👍 thanks

PR #660, opened 2022-04-28
Adjustments to `start` Options

eric1234:
Could not get the test suite to pass. Ran into two issues due to options (or lack of options) being passed to the `start` function. Rather than adjust the test suite to pass the necessary option I thought it would be better to adjust the API to have defaults so the options are not necessary.

Spawn Options Stdio
================

This defaults to [`pipe`](https://nodejs.org/api/child_process.html#optionsstdio). With this default the behavior I was getting is that the `checkStarted` was always failing. This was obviously not an issue at some point. I am guessing that the selenium jar file changed it's behavior so that it didn't fully boot up until some stdio activity occurred and since this library is not interacting with that pipe it remained not booted.

I noticed the `bin/selenium-standalone` script did not have a problem booting the server. After a bit of debugging determined it is because [it sets the spawnOptions.stdio to `inherit`](https://github.com/webdriverio/selenium-standalone/blob/main/bin/selenium-standalone#L125-L127). This allows that sub-process to not be blocked by stdio.

I considered also defaulting spawnOptions.stdio to 'inhert' with the API but I'm not sure writing on top of the parent's STDIO is the right default behavior. It would seem being quite and squashing stdio (sending to /dev/null) is a better default so I set it to
`ignore`. Obviously if this is not the right behavior for a specific application they can always override.

Add `--detect-drivers false` Option
==========================

Once the above fix to the boot process was made I was still getting a problem with one test, "should start with the given drivers", where no driver are being specified.

When this is run the selenium-standalone jar was giving me the following error:

```
10:56:47.730 INFO [LoggingOptions.configureLogEncoding] - Using the system default encoding
10:56:47.733 INFO [OpenTelemetryTracer.createTracer] - Using OpenTelemetry for tracing
10:56:48.170 INFO [NodeOptions.getSessionFactories] - Detected 8 available processors
10:56:48.181 INFO [NodeOptions.discoverDrivers] - Discovered 0 driver(s)
10:56:48.187 WARN [NodeOptions.addDetectedDrivers] - No drivers have been configured or have been found on PATH
java.lang.reflect.InvocationTargetException
 at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
 at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
 at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #670, opened 2022-05-06
Fix conditional driver loading

eric1234:
This is a follow-up to #660.

The "should start with the given drivers" test was not passing for me. I believe this is because `--detect-drivers` is auto-loading drivers it can find on some peoples system but on my system there are no drivers in the PATH so it causes an error by the JAR file. Either that or it is failing for everybody and the JAR file changed behavior recently to throw an error when no drivers can be found and `--detect-drivers` is enabled.

In #660, I thought `--detect-drivers` only did auto-detection and since we were specifying the path to all drivers it should not be needed. By setting that to false it caused the tests to pass. Unfortunately as @wyhmichael correctly pointed out that option will stop all loading of drivers not just the ones auto-detected. We should be using the `-I` option to indicate what drivers to load.

This PR does 3 things:

* 6503855 reverts the revert of my original work. I wanted to do this because that work had other fixes we want to keep (stdio fixes) as well as I would like the wrong code to initially be in the system so we can update the test suite to detect the problem I created.
* 7f789b3 makes that update to the test suite to detect the problem. Instead of examining the arguments to see what was asked to be loaded it examines the Selenium output to see what was actually loaded. By making this change my wrong code is no longer passing.
* 3fcbe26 wraps this up by removing my wrong code and replacing it with the `-I` arguments that I think make the right code.

Commit messages have a bit more info but hopefully this should work better than my last attempt.

### DIFF
--- a/lib/start.js
+++ b/lib/start.js
@@ -42,8 +42,6 @@ async function start(_opts) {
     !hasParam(opts.seleniumArgs, 'standalone') &&
     !hasParam(opts.seleniumArgs, 'distributor')
   ) {
-    opts.seleniumArgs.unshift('false');
-    opts.seleniumArgs.unshift('--detect-drivers');
     opts.seleniumArgs.unshift('standalone');
   }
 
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):
Another excellent PR, well done 👍

bmcfeely:
Hi @eric1234! I'm still digging in, but I think this change might be preventing safari from opening. I'm still sorting this out, just a heads up. I seeing that when I remove the -I args from the command line, then selenium-standalone will open up a safari session, but with those `-I` args as they are now, selenium-standalone seems to silently do nothing.

bmcfeely:
@christian-bromann @eric1234 Thanks all! I'll get a PR up in the next hour or so.

bmcfeely:
Done! The code comment here may be a little wordy, I can edit if it's too much. 😄  https://github.com/webdriverio/selenium-standalone/pull/675

christian-bromann:
Sweet, thanks for taking a stab at it 👏  Let's continue convo in the PR.

PR #675, opened 2022-05-13
feat: Allow Safari to be launched

bmcfeely:
#670 changed the launching of `selenium-server.jar` to only allow certain browsers. Although there is no driver that needs downloading for Safari (apparently SafariDriver is available on all Macs), we should still allow Safari to be launched.

Testing:
- Verified that I can now request `browserName: 'safari'` and that Safari loads on my Mac

### DIFF
--- a/lib/start.js
+++ b/lib/start.js
@@ -122,6 +122,16 @@ async function start(_opts) {
     delete fsPaths.chromiumedge;
   }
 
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
Nice 👍 

Thank you!

PR #710, opened 2022-09-30
Update compute-download-urls.js

Ghop11:
updated platform to handle M1 MacBooks new chromedriver zip file naming convention.  "mac_arm64"


christian-bromann:
@Ghop11 thanks for raising the PR, two things:

- mind signing the CLA?
- does this also work if someone would like to download an older version of Chromedriver?

christian-bromann:
Fixed by #711

thiago-luppi-td:
hey guys, Is this fixed because I'm dealing with this error right now in November.

PR #716, opened 2022-10-07
Use Mac arm64 EdgeDriver when available

ianrenauld:
I had an issue yesterday downloading the latest Chromium EdgeDriver on my M1 MacBook Air from a fresh install. It turns out that the latest EdgeDriver version had only been released for Linux. Seemingly Microsoft releases each different OS versions one at a time, not updating everything all at once. Even this morning, I see that selenium-standalone cannot download the latest version, it uses the fallback version 96.0.1054.34 even though I can manually download the latest versions from https://developer.microsoft.com/en-us/microsoft-edge/tools/webdriver/

I do not have a solution for that (maybe I should enter a bug about the fallback mechanism being suboptimal?), but while investigating, I realized that selenium-standalone is not downloading the arm64 version of EdgeDriver for arm64 Macs as it does for Chrome.

Here is a PR to solve this. I noticed that the default-downloads-test.js Unit Test had no section for Chromium Edge, but trying to add one with the piece by piece update strategy from Microsoft was very difficult (due to the issue above about always triggering the fallback mechanism). In the end, I gave up adding a test for Chromium Edge in default-downloads-test.js, maybe that's the reason why there wasn't one to start with as it seems very difficult to make it work consistently?

If you have any guidance about improving that fallback process (and maybe we can't, I guess. It all depends on how Microsoft handles their release of EdgeDriver) maybe I can take a stab at it?

One last thing, it seems that for each major version of Chromium EdgeDriver, the early releases don't necessarily have a arm64/"m1" Mac release, but the later releases under that same major version do.  So the cutoff of when to try to grab the arm64/"m1" release was pretty much arbitrary. I decided that major version 105 and up is ok, but if you have a better idea, I am all ears!

### DIFF
--- a/lib/compute-download-urls.js
+++ b/lib/compute-download-urls.js
@@ -101,7 +101,7 @@ async function computeDownloadUrls(opts) {
         urls.chromiumedge,
         opts.drivers.chromiumedge.baseURL,
         opts.drivers.chromiumedge.version,
-        getChromiumEdgeDriverArchitecture(opts.drivers.chromiumedge.arch)
+        getChromiumEdgeDriverArchitecture(opts.drivers.chromiumedge.arch, opts.drivers.chromiumedge.version)
       );
   }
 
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):
Awesome, thanks 👍

christian-bromann:
I've been wanting to consolidate the driver management for a long time. We are maintaining 4 driver services in the community channel and this project which basically also does the same for all drivers at once. The Selenium project is working on a webdriver manager written in rust which could be used as a replacement (as a NPM package using WASM). That said, very appreciate patches like this but I am sceptic about bigger dev efforts for this project due to these developments.

PR #744, opened 2022-12-30
as selenium-standalone still support node": ">=12.0.0 prevent install…

udarrr:
as selenium-standalone still support node": ">=12.0.0 prevent install not supportive fs-extra


christian-bromann (approved):
👍

christian-bromann (commented):
Please also update the lock file

PR #802, opened 2023-07-04
use opt.drivers.chome.arch to determine arch on mac

### LINKED ISSUE #797: Chrome driver architecture setting is ignored and can result in downloading x64 binary into arm64 folder
## Expected Behavior

When specifying the arm64 architecture for chrome drivers, selenium-standalone should download the arm64 architecture, even if running in an x64 process.

## Current Behavior

`compute-download-urls.js` ignores the `wantedArchitecture` argument. This means I can specify that I want the `arm64` architecture, but if `process.arch` says I'm running in x64, then it'll download the x64 binary.

https://github.com/webdriverio/selenium-standalone/blob/a40cb25a53f05fabcd6b8c248c2c7a4d39ae61c4/lib/compute-download-urls.js#L111C38-L111C56

However, `compute-fs-paths.js` correctly saves the driver into `.selenium/chromedriver/latest-arm64`, meaning that an x64 binary gets downloaded into a folder labelled as if it contains an arm64 binary.

For me, this was leading to difficult to diagnose performance problems, because the chromedriver binary kept running the Intel translation layer on my M1 macbook.

## Steps to Reproduce

WDIO services config:

```
  services: [["selenium-standalone", {
    installArgs: {drivers: {chrome: {arch: 'arm64'}}},
    args: {drivers: {chrome: {arch: 'arm64'}}},
  }]],
```

when running in a Rosetta terminal (x64 architecture) on an M1 mac.

Alternatively:

```
npx selenium-standalone install --singleDriverInstall=chrome --drivers.chrome.arch=arm64
```

## Logs

```
----------
----------

---
selenium install:
from: https://github.com/SeleniumHQ/selenium/releases/download/selenium-4.4.0/selenium-server-4.4.0.jar
to: /Users/gpaciga/selenium-standalone-arch/node_modules/selenium-standalone/.selenium/selenium-server/4.4.0/selenium-server.jar
---
chrome install:
from: https://chromedriver.storage.googleapis.com/114.0.5735.90/chromedriver_mac64.zip
to: /Users/gpaciga/selenium-standalone-arch/node_modules/selenium-standalone/.selenium/chromedriver/latest-arm64/chromedriver


-----
selenium-standalone installation finished
-----
```

## Your Environment
* Version of `selenium-standalone` that you are using: 8.11, but dates back to at least 7.19.1
* Is there another tool calling `selenium-standalone` on your behalf: WebdriverIO
* System/platform: Macbook pro M1 
* Run install command within an x64 terminal

gpaciga:
resolves #797

### DIFF
--- a/lib/compute-download-urls.js
+++ b/lib/compute-download-urls.js
@@ -194,7 +194,8 @@ function getChromeDriverArchitecture(wantedArchitecture, version) {
   if (process.platform === 'linux') {
     platform = 'linux64';
   } else if (process.platform === 'darwin') {
-    if (process.arch === 'arm64') {
+    const arch = wantedArchitecture || process.arch;
+    if (arch === 'arm64') {
       const [major] = version.split('.');
       platform = parseInt(major, 10) > 105 ? 'mac_arm64' : 'mac64_m1';
     } else {
--- a/test/compute-download-urls-test.js
+++ b/test/compute-download-urls-test.js
@@ -232,6 +232,36 @@ describe('compute-download-urls', () => {
         const actual = await computeDownloadUrls(opts);
         assert.strictEqual(actual.chrome, 'https://localhost/106.0.5249.61/chromedriver_mac_arm64.zip');
       });
+
+      it('Use `mac_arm64` on x64 arch when requested arch is arm64', async () => {
+        opts.drivers.chrome = {
+          baseURL: 'https://localhost',
+          version: '106.0.5249.61',
+          arch: 'arm64',
+        };
+
+        Object.defineProperty(process, 'arch', {
+          value: 'x64',
+        });
+
+        const actual = await computeDownloadUrls(opts);
+        assert.strictEqual(actual.chrome, 'https://localhost/106.0.5249.61/chromedriver_mac_arm64.zip');
+      });
+
+      it('Use `mac64` on arm64 arch when requested arch is x64', async () => {
+        opts.drivers.chrome = {
+          baseURL: 'https://localhost',
+          version: '106.0.5249.61',
+          arch: 'x64',
+        };
+
+        Object.defineProperty(process, 'arch', {
+          value: 'arm64',
+        });
+
+        const actual = await computeDownloadUrls(opts);
+        assert.strictEqual(actual.chrome, 'https://localhost/106.0.5249.61/chromedriver_mac64.zip');
+      });
     });
 
     describe('win', () => {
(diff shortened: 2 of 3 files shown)


christian-bromann:
Would you mind signing the CLA?

tehhowch:
@gpaciga see above comment

gpaciga:
@tehhowch already done a couple weeks ago

tehhowch:
@christian-bromann bump

tehhowch:
@gpaciga thoughts on the failing CI checks?

gpaciga:
@tehhowch  the URLs that are expected for the nightly build are not accurate; it uses `https://github.com/SeleniumHQ/selenium/releases/download/selenium-nightly/selenium-server-standalone-nightly.jar` but the actual URL for the nightly build (at least today) is `https://github.com/SeleniumHQ/selenium/releases/download/nightly/selenium-server-4.11.0.jar`.

But, this is an existing issue unrelated to this PR. The same problem exists on main.

christian-bromann:
Did this get potentially fixed with #817?

gpaciga:
#817 certainly changes things but it does not fix this issue.

There's a new if condition on line 119 that seems to be redundant with the one on line 137

It looks like the new branch starting on line 70 means that if the user specifies `opts.drivers.chrome.version`, it again uses _different_ faulty logic that still ignores the user's stated architecture preference (instead relying purely on `os.arch()`)

Just from trying to understand the code, it looks like there are at least 3 scenarios:
- when opts.drivers.chrome.version is "latest"
- when opts.drivers.chrome.version is <= 114
- when opts.drivers.chrome.version is > 114

but this PR currently only address the <=114 case.

I've rebased on main and did my best to add tests to illustrate the bug in those new scenarios, but don't know how I should proceed from here to fix the code that was added by #817.

gpaciga:
It seems like this should all be moot since this service is not required past WDIO 8.14 anyway.

@christian-bromann can you point me to the new logic where WDIO handles getting chromedriver internally? I wasn't able to find it, but it appears to still be incorrectly running with x64 architecture in my setup.

christian-bromann:
> can you point me to the new logic where WDIO handles getting chromedriver internally?

Please see https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-utils/src/driver/utils.ts#L75-L201 .. we are using the `@puppeteer/browser` package to determine the right version to download.

udarrr:
here is fix https://github.com/webdriverio/selenium-standalone/pull/826

christian-bromann:
Thanks @udarrr , merged and released the fix.

PR #817, opened 2023-08-08
chrome for testing feature

udarrr:


udarrr:
@christian-bromann Could you check it ?

package.json
```diff
@@ -1,15 +1,15 @@
 {
   "name": "selenium-standalone",
-  "version": "8.3.0",
+  "version": "9.0.0",
```
christian-bromann, line 3: We should revert this as this project has its own pipeline for releasing and bumping the version.

christian-bromann (commented):
One comment, but otherwise looks good to me 👍 
great job!

package.json
udarrr: to the 8.3.0 ?

christian-bromann: yes

udarrr: done :)

christian-bromann (approved):
Fantastic work 👏 
Thank you!

tehhowch (commented):
Did adding support for "Chrome for Testing" _require_ also modifying the node version requirements? I think the only "reason" is because a totally new dependency was added (on the [`@puppeteers/browsers`](https://www.npmjs.com/package/@puppeteer/browsers) package). A backport to v8.x that is still compatible with node 12 would be useful to lots of folks that cannot yet use Webdriver v8 / Node 16 (e.g. because sync to async migrations are still ongoing, or other internal reasons).

PR #819, opened 2023-08-08
the latest for old versions of chromedriver

udarrr:


udarrr:
Hi @christian-bromann could you check it? Here is simple fix for backward combability with old version of chromedriver (useful for electron)

christian-bromann (approved):
Nice, lgtm 👍

PR #820, opened 2023-08-09
added return values for install

udarrr:


udarrr:
@christian-bromann  Could you check it? here is just minor update for install method

christian-bromann (approved):
LGTM 👍

PR #836, opened 2023-08-25
backward compatibility

udarrr:


udarrr:
Hi @christian-bromann  Generally the contribution with backward compatibility with previous versions of nodes. Previously we've used @puppeteer/browsers but just for one action in a way to obtain the latest version of chromedriver, i've replace it to api and have suspend @puppeteer/browsers. It has been tested since 12 to 20 node. And i'm not convinced what kind of version it should be)

christian-bromann (approved):
Fantastic work 🎉 

I will go ahead and just make a patch release.

Issue #599, opened 2021-12-06
[🐛 Bug]: Selenium does not "frees" the port


joaodanielopes:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

7.16.6

### Node.js Version

v16.13.0

### Mode

WDIO Testrunner

### What happened?

I am able to run the test on the first time.
However, on the next runs, I got the error:

**ERROR @wdio/selenium-standalone-service: Error: Port 4444 is already in use.**

To "free" the port I need to manually kill the process running in the Windows command prompt:
netstat -aof | findstr :4444
taskkill /PID <pid> /F

### What is your expected behavior?

After the run, selenium stops and "frees" the ports.

### How to reproduce the bug.

Run a project with the following settings:

- java version "17.0.1" 2021-10-19 LTS

- package.json:

`{
    "name": "name",
    "version": "1.0.0",
    "description": "description",
    "scripts": {
        "build": "tsc",
        "test": "wdio wdio.conf.ts"
    },
    "dependencies": {
        "@cucumber/cucumber": "^7.3.1",
        "@types/chai": "^4.2.22",
        "@types/node": "^16.11.11",
        "@wdio/cli": "7.16.6",
        "@wdio/cucumber-framework": "7.16.6",
        "@wdio/local-runner": "7.16.6",
        "@wdio/selenium-standalone-service": "7.16.6",
        "chai": "4.3.4",
        "multiple-cucumber-html-reporter": "1.18.0",
        "ts-node": "10.4.0",
        "tsconfig-paths": "3.12.0",
        "typescript": "4.5.2",
        "wdio-cucumberjs-json-reporter": "3.0.0",
        "webdriverio": "7.16.6"
    }
}`

- wdio.conf.ts:

`{
    runner: "local",
    specs: ["./src/features/**/*.feature"],
    exclude: [],
    maxInstances: 1,
    capabilities: [
        {
            browserName: "chrome",
            'goog:chromeOptions': {
                args: [
                    '--disable-infobars',
                    '--start-maximized'
                ]
            },
        },
    ],
    logLevel: "warn",
    bail: 0,
    baseUrl: "https://www.google.com",
    waitforTimeout: 20000,
    connectionRetryTimeout: 60000,
    connectionRetryCount: 3,
    services: [
        [
[...]

PR #838, opened 2023-08-26
Process killer feature

udarrr:
@christian-bromann Generally it's a good feature when we can check and kill selenium processes by port :4444 and drivers names before starting.

Tests added

Bug is here https://github.com/webdriverio/selenium-standalone/issues/599

And I'm not convinced but maybe it have sense to be switched on in config by default. Currently it's not switched on.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -111,7 +111,7 @@ By default all drivers are loaded, you only control and change the versions or a
 
 `opts.requestOpts` can be any valid [`got` options object](https://www.npmjs.com/package/got#proxies). You can use this for example to set a timeout.
 
-`opts.processKiller` set port of selenium server for killing, could be ;ile `4444` or `:4444`.
+`opts.processKiller` set it like true for killing selenium server port.
 
 returns `Promise<ChildProcess>`
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -111,7 +111,7 @@ By default all drivers are loaded, you only control and change the versions or a
 
 `opts.requestOpts` can be any valid [`got` options object](https://www.npmjs.com/package/got#proxies). You can use this for example to set a timeout.
 
-`opts.processKiller` set it like true for killing selenium server port.
+`opts.processKiller` set for that truthy value, for killing selenium server port.
 
 returns `Promise<ChildProcess>`
 
@@ -123,7 +123,7 @@ If you're getting this error, it means that you didn't shut down the server succ
 pkill -f selenium-standalone
 ```
 
-or use `opts.processKiller` in config
+or use truthy `opts.processKiller` in config
 
 ## Set `selenium-standalone` Version as NodeJS environment parameter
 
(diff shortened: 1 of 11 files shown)


udarrr:
@christian-bromann Generally it's a good feature when we can check and kill selenium process by port :4444 with hanged drivers processes before starting.

docs/API.md
```diff
@@ -111,6 +111,8 @@ By default all drivers are loaded, you only control and change the versions or a
 
 `opts.requestOpts` can be any valid [`got` options object](https://www.npmjs.com/package/got#proxies). You can use this for example to set a timeout.
 
+`opts.processKiller` set port of selenium server for killing, could be ;ile `4444` or `:4444`.
```
christian-bromann, line 114: ;ile?

christian-bromann (commented):
I am not sure how useful this would be for a user given they might not know on which port the driver was run. Can't we just improve how we kill the process and try another attempt if we we detect that the port is still occupied?

I think it would be a better approach to actually guarantee to free up the port.

udarrr:
> I am not sure how useful this would be for a user given they might not know on which port the driver was run. Can't we just improve how we kill the process and try another attempt if we we detect that the port is still occupied?
> 
> I think it would be a better approach to actually guarantee to free up the port.

@christian-bromann Yes agreed it has sense, not necessary kill it if port is free already, i've added changes

docs/API.md
```diff
@@ -111,6 +111,8 @@ By default all drivers are loaded, you only control and change the versions or a
 
 `opts.requestOpts` can be any valid [`got` options object](https://www.npmjs.com/package/got#proxies). You can use this for example to set a timeout.
 
+`opts.processKiller` set for that truthy value, for killing selenium server port.
```
christian-bromann, line 114:
```suggestion
`opts.processKiller` set to a truthy value, for killing selenium server port.
```

docs/API.md
```diff
@@ -121,6 +123,8 @@ If you're getting this error, it means that you didn't shut down the server succ
 pkill -f selenium-standalone
 ```
 
+or use truthy `opts.processKiller` in config
```
christian-bromann, line 126:
```suggestion
or use `opts.processKiller` in config
```

docs/CLI.md
```diff
@@ -49,6 +49,9 @@ selenium-standalone start --singleDriverStart=chrome
 selenium-standalone install --config=/path/to/config.json
 selenium-standalone start --config=./config/seleniumConfig.js
 
+# killing process before starting
+selenium-standalone start --processKiller=true
```
christian-bromann, line 53: A user would ask why they would need to set this given it feels like the expected behavior is that the selenium server or driver shuts down completely when the process is triggered to get shutdown.

christian-bromann (commented):
It still is a bit confusing to me why I would need to set `opts.processKiller` because as mentioned in the comment as a user I would expect that this package is able to shut down the selenium server process and clear the port automatically.

christian-bromann (approved):
Ok, sgtm 👍

PR #841, opened 2023-08-28
killing by default

udarrr:
like continuation https://github.com/webdriverio/selenium-standalone/pull/838 

@christian-bromann I think it worth to try your point of view. Believe everything should work.


udarrr:
i've changed tests, currently everything is ok

PR #845, opened 2023-09-17
chore/update code format, remove one place package usage

potapovDim:
Removed fs-extra package from dependencies since it was used in one place, 
nodejs has fs.existsSync, in this case it is not a long check.
Removed global context reference from functions body and some code style alignment


christian-bromann (approved):
Nice 👌

PR #846, opened 2023-09-18
chore/update code format, replace declaration with expressions

potapovDim:


udarrr:
Hi @potapovDim  I am just curious about why did you do that ? What's wrong with function declaration?

potapovDim:
@udarrr nothing wrong with fn declaration, just to have a common code approach format across library i started to update it to expression format also expression could not be hosted in a way as declarations so better for reading, would be also good to add jsdoc comments and some lint format/prettier

christian-bromann:
@potapovDim maybe it makes sense to introduce Prettier or EsLint to ensure that the next contributor doesn't switch this around again. I don't mind which code style we go with as long as we ensure we stick with it using these tools.

potapovDim:
@christian-bromann i can do that, can we merge this one and i will proceed with that next steps ?

christian-bromann:
> or do you want me to add this one as a part current PR ?

Let's add it to this PR if you don't mind

udarrr:
> > or do you want me to add this one as a part current PR ?
> 
> Let's add it to this PR if you don't mind

Oh That's what it was... yes there is a invitation thank you @christian-bromann ! I do my best :)

PR #862, opened 2023-12-13
[fix] chromiumedge driver doesn't install on arm64 when user specifies "latest" for version

jdpolicano:
This fixes a small error when installing the chromium edge driver on the arm64 architecture. When the user selects "latest" for version, we were not passing the version info to the getChromiumEdgeDriverArchitectureOld function from chromiumEdgeBundleAvailable. This caused the function to fail silently, which would revert to the fallback version even if a later version was available.


christian-bromann (approved):
LGTM 👍

jdpolicano:
Thanks!

PR #875, opened 2024-02-16
fix: Update chromedriver download base url

just-boris:
Current url does not work for latest drivers: https://github.com/GoogleChromeLabs/chrome-for-testing/pull/108

```
404 https://edgedl.me.gvt1.com/edgedl/chrome/chrome-for-testing/121.0.6167.184/mac-arm64/chromedriver-mac-arm64.zip
Response code 404 (Not Found)
```

Works well with the updated urls


christian-bromann (approved):
👍

PR #914, opened 2024-06-24
added proxy complience

udarrr:
1. Added proxy for http, https requests for cases when the tool behind corporate proxy and system env HTTP,HTTPS were specified
2. Dropped support node12 as http and https are node14 tools

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node-version: [12, 14, 16, 18, 20]
+        node-version: [14, 16, 18, 20]
         java-version: [11]
         include:
           - os: macos-latest
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,6 @@
+# 9.6.0 (2023-10-21)
+* added proxy complience
+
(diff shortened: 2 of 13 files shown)


udarrr:
@christian-bromann do you mind review it and if ok then trigger release ?

.github/workflows/test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node-version: [12, 14, 16, 18, 20]
```
christian-bromann, line 14: If we drop support for Node.js v12 we have to do a major release

HISTORY.md
```diff
@@ -1,3 +1,6 @@
+# 9.6.0 (2023-10-21)
```
christian-bromann, line 1: if we do a major release this should be `10.0.0`
```suggestion
# 10.0.0 (2024-06-24)
```

christian-bromann (changes requested):
One question on the Node.js v12 drop

.github/workflows/test.yml
udarrr: yep it makes sense

udarrr:
> One question on the Node.js v12 drop

done

christian-bromann (approved):
👍

